PR #694, opened 2013-08-31
[cookbook] Dynamic views based on model properties

ianpetzer:


trek:

PR #696, opened 2013-08-31
[cookbook] "Loading Data from a server" updated to match Recipe guidelines

mavilein:
I moved the discussion part into the solution and omitted a discussion, since i have nothing interesting to write about it.


trek:

PR #704, opened 2013-08-31
Contributing to the getting started guide

trek:


trek:
There is a new GSG, so this should just be closed.

PR #830, opened 2013-09-07
[cookbook] Added cookbook for notifying users of updated application versions

robmonie:


trek:

PR #1285, opened 2014-02-10
Update show-when-all-todos-are-complete.md

lfalin:
Updated to remove deprecated everyProperty method and obscure double not operator. 
Also changed property dependency to depend upon the completed property, which matches what the hasCompleted property looks like from the previous step in the guide.

### DIFF
--- a/source/guides/getting-started/show-when-all-todos-are-complete.md
+++ b/source/guides/getting-started/show-when-all-todos-are-complete.md
@@ -16,8 +16,8 @@ In `js/controllers/todos_controller.js` implement the matching `allAreDone` prop
 ```javascript
 // ... additional lines truncated for brevity ...
 allAreDone: function(key, value) {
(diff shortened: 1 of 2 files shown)


aitherios:
Wouldn't be possible to use the remaining property?

Instead of:

```javascript
allAreDone: function(key, value) {
  if (value === undefined) {
    return !!this.get('length') && this.isEvery('isCompleted', true);
  } else {
    this.setEach('isCompleted', value);
    this.invoke('save');
    return value;
  }
}.property('@each.isCompleted')
```

To be like:

```javascript
allAreDone: function(key, value){
  if (value === undefined) {
    return this.get('remaining') === 0;
  } else {
    this.setEach('isCompleted', value);
    this.invoke('save');
    return value
  }
}.property('remaining')
```

trek:
The new GSG will not be based on TodoMVC any longer.

PR #1367, opened 2014-03-13
fixing getting-started bug when editing todo

izhan:
Editing todos throws an error whenever the change is confirmed by using a newline (Enter).  In particular, it throws the error

`Uncaught Error: Attempted to handle event 'willCommit' on <Todos.Todo:ember270:2> while in state root.loaded.updated.inFlight. `

For the final Todo (Replacing the Fixture Adapter with Another Adapter), it throws the error

`Uncaught Error: Attempted to handle event 'willCommit' on <Todos.Todo:ember272:oumdk> while in state root.deleted.saved. `

when you attempt to delete a todo by editing the item to be blank.  This error throws future errors in some cases such as `Assertion failed: calling set on destroyed object `, making the todo list unable to function.  Note that this bug does not occur when one confirms an edit by clicking somewhere out of the textfield.

This bug occurs because when newline triggers an edit or removal of a todo, `acceptChanges` is called a second time due to the textfield being focused out.  This may or may not be intended behavior -- PR fixes this and all JSBins by checking whether it is still in edit mode.

### DIFF
--- a/source/guides/getting-ember/index.md
+++ b/source/guides/getting-ember/index.md
@@ -24,7 +24,6 @@ Adding Ember to your application with Bower is easy simply run `bower install em
 		"ember-data": "~1.0.0-beta.4"
 	}
 }
-
 ```
 
 ##RubyGems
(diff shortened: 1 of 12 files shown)


trek:

PR #1478, opened 2014-04-29
Fix description of records

eneiand:
A record is not defined by having obtained its data from a server. Also I think the intent was to say that the id is 1 and the name is steve-buscemi, otherwise it is confusing.


trek:

PR #1533, opened 2014-05-21
Update show-when-all-todos-are-complete.md

rayjcwu:
change it to be identical to the demo


trek:

PR #1544, opened 2014-05-28
Drag and Drop components cookbook example

kalmanh:


trek:

PR #1566, opened 2014-06-15
Document how to find the end of the loading event.

wbyoung:
I found it difficult to discover how to actually be notified of when the end of a transition occurred after receiving the `loading` event. I believe this is the correct way, and if so, I think it'd be helpful for others to see this example.


trek:

wbyoung:
Thanks, opened emberjs/guides#119.

PR #1584, opened 2014-06-30
There are Ember-Handlebars plugins for build tools out there

silvenon:


trek:

PR #1634, opened 2014-07-31
Add documentation to for the ActiveModelAdapter

opsb:
This PR adds some documentation for the ActiveModelAdapter. It includes instructions for setting up both rails and ember. It also talks about how you specify the serialization of relationships on the ember side.

This is a first draft really so any suggestions for things to include are very welcome.

The docs also assume that the PR https://github.com/emberjs/data/pull/2138 will be accepted(it changes how associations are embedded to match the rails conventions - mainly that relationships append _attributes to their name).


trek:

PR #1664, opened 2014-08-18
Create hierarchical_navigation_with_nested_templates

Lepetere:
I had the problem that I couldn't find a real example for this basic use case. So I'm adding it here after I figured out how to do this. I ran into some problems concerning the naming of things, so I think this might be a great help to people new to Ember. But since I'm not an expert yet, feel free to correct my suggestion if I still broke any 'Ember rules'.


trek:

PR #1689, opened 2014-09-01
Update dependency-injection-and-service-lookup.md

alexmasita:
Not an expert so ignore the patch if totally stupid.  No offence will be taken.  Had trouble wrapping around the dependency injection language in the first paragraph.  Have tried to clarify it in a way others like me may find understandable.


trek:

PR #1722, opened 2014-09-23
Add nested this.route to router docs

cavneb:
Added section describing usage of nested `this.route`

screen shot 2014-09-23 at 1 00 30 pm


trek:

PR #1809, opened 2014-11-14
Recomend using ember cli in getting ember

raytiley:
The 2.0 RFC recomends moving to ember-cli asap, its probably time to start reflecting that in the guides


trek:
This will be addressed new versioned guides.

PR #1812, opened 2014-11-15
Aligns tutorial text with working code

adsteel:
Tutorial code wasn't working. This working code was taken directly from the JSBin example on the tutorial page.


trek:

PR #1835, opened 2014-11-21
Match docs to code diffs

kclowes:
When I was going through the tutorials, I had to follow the code in the Github diffs, instead of the code in the tutorials. There is already a PR in for 'displaying-model-data', but the rest of the files in here did not match the diffs on Github, so I went ahead and changed all the discrepancies I saw so that it was all in one place. Thanks!


trek:

PR #1847, opened 2014-11-29
First pass at improving Routing DSL `route` vs `resource` docs

eoinkelly:
I noticed that the router guide was not very clear about the difference between `route` and `resource` in the router DSL especially given that `route` can have nested routes since 1.7.

First pass only, happy to make changes :-)


trek:

PR #1849, opened 2014-12-01
[Cookbook] adding a problem to add dynamic input elements on the page to...

ankeetmaini:
... add a row to a table or a new item in a list


trek:

PR #1857, opened 2014-12-09
Remove TodoController to mark completed/incompleted

matthewrankin:
As of Ember 1.8.1 (possibly 1.8), it appears you no longer need to setup the TodoController in order to mark an item complete or incomplete. Once the input checkbox is tied to `todo.isCompleted`, marking an item complete or incomplete just works.

The fact that Ember makes this work automatically should probably be discussed, but I don't feel that I have enough Ember knowledge to write that.

Additionally, the links to the jsbin live preview and the changes in `diff` format still need to be updated before this branch is merged.


trek:

PR #1898, opened 2014-12-30
Remove incorrect instruction to append with `Id`

artfuldodger:
The model name should no longer be appended with `Id`, and this is reflected in the example code but not in the text.


trek:

PR #1903, opened 2015-01-02
To support windows, no need so long README, just add to Gemfile is enough.

Eric-Guo:
also do some gem upgrade.


Gemfile
```diff
@@ -27,3 +27,6 @@ group :test do
   gem "poltergeist"
 end
 
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+gem 'tzinfo-data', platforms: [:mingw, :mswin]
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```
fivetanley: looks great, i wonder why these two lines don't use the same platform directive though. seems like they should both use the same conditional..?

Eric-Guo: Good question and show ruby have having many way to do one things. :smile: 

Any way, change to mingw and mswin platform only now

fivetanley:
thanks!

PR #1905, opened 2015-01-03
instructions say to remove wrong class

thegreatmichael:
(perhaps I missed a step, but) at this point in the tutorial, the class on the link is `selected` not `active`


trek:

PR #1910, opened 2015-01-04
[Ember Users] adding Netural to the list of Ember users

ddprrt:
We use Ember for several projects, the latest one being [Reviermanager](http://webapp.reviermanager.at)


kmiller915:
Email received. cc @wifelette

fivetanley:
thanks!

Issue #1913, opened 2015-01-05
Deter beginners from creating Views over Components


davidkovsky:
I'm having a great time with Ember. The guides are very helpful. Unfortunately, I only recently realized that I should have been creating Components instead of Views in my own project. I will create a Pull Request that simply adds a note to the top of source/guides/views/index.md in the hopes that it helps someone else in the future.

davidkovsky:
https://github.com/emberjs/website/pull/1914

PR #1914, opened 2015-01-05
deters beginners from choosing Views over Components

davidkovsky:
This is my suggested short-term resolution for [issue 1913](https://github.com/emberjs/website/issues/1913).


trek:
There are new guides in the works. We'll leave Views around for older versions.

PR #1915, opened 2015-01-05
Cookbook create flash messages

ianwdunlop:
Cookbook recipe for creating flash message alerts. Uses an ember data model to create the flash message and an array controller to keep track of them. Each flash message has a view and a template along with action handler to remove them.


trek:

PR #1916, opened 2015-01-05
Fix if statement in transition documentation

calvinlough:


source/guides/routing/preventing-and-retrying-transitions.md
```diff
@@ -54,7 +54,7 @@ destination routes to abort attempted transitions.
 ```js
 App.DiscoRoute = Ember.Route.extend({
   beforeModel: function(transition) {
-    if (new Date() < new Date("January 1, 1980")) {
+    if (new Date() > new Date("January 1, 1980")) {
       alert("Sorry, you need a time machine to enter this route.");
```
calvinlough: But if your date is before 1980, you get an error saying you need a time machine. Shouldn't it be the other way around?

beerlington: I'm pretty sure @calvinlough is correct on the change if the intent is to abort the transition for users who enter the route after 1980.

trek:

PR #1924, opened 2015-01-09
Add Duda as an ember-user

ramybenaroya:


twokul:
cc @wifelette

wifelette:
Email received, cc @kmiller915

PR #1927, opened 2015-01-11
Change transitionTo in favor of transitionToRoute

amedr:
transitionTo is about to be deprecated.


trek:

PR #1931, opened 2015-01-12
Remove a superfluous 'that'

kxprim:
The word 'that' did not make sense in the sentence, I therefore removed it.


fivetanley:
you should probably rebase this off your other change otherwise it'll be re-added in this commit it looks like.

PR #1934, opened 2015-01-12
Fix DOM spec organization

domenic:
The DOM Standard https://dom.spec.whatwg.org/ is maintained by the WHATWG, not the W3C.


fivetanley:
Thanks!

domenic:
Thank *you* good sir!

PR #1936, opened 2015-01-13
Include the auto location option

artfuldodger:
Being able to support `history` when available with a fallback seems like it's worth calling out.


trek:

Issue #1881, opened 2014-12-18
Embed jsbin examples broken


bbonamin:
All jsbin examples in the Testing guide are broken. 

It seems that Ember is not included as a library, because editing those examples and including Ember fixes them.

![captura de pantalla 2014-12-17 21 06 56](https://cloud.githubusercontent.com/assets/713787/5481485/aa57cafa-8630-11e4-8bc9-c8b982d67186.png)

patocallaghan:
I've found updating the Handlebars version to the latest version fixes these bins

```
<script src="http://builds.handlebarsjs.com.s3.amazonaws.com/handlebars-v2.0.0.js"></script>
```

bjornstar:
These tests are still broken on the live site, any chance this code can get pushed live?

ryanlitalien:
I just looked into this.  Some JS Bin examples are pointing to

```<script src="http://builds.emberjs.com/ember-latest.js"></script>```

Visiting that URL results in a 404. It looks like they moved the latest builds directory to: http://emberjs.com/builds/#/canary

Example JS Bin:
Live site - http://emberjs.com/guides/components/passing-properties-to-a-component/
Source - https://github.com/emberjs/website/blob/master/source/guides/components/passing-properties-to-a-component.md
JS Bin - http://jsbin.com/japiv/2/embed?live

Quite a few JS Bin's will need to be updated.  Not sure of the proper ember build # to reference and the process for updating the source.

trek:
We've migrated to versioned guides at https://github.com/emberjs/guides and no longer use JSBin

supairish:
@trek Can you clarify (beginner here).  The v1.10.0 guides still have JSBins in them, for example http://guides.emberjs.com/v1.10.0/components/passing-properties-to-a-component/

Are you saying those are going away like they are in the v1.11 guide?  Or are the v1.10 guide JSBins just a no fix at this point?

trek:
Each revision of the guides is stored as a snapshot. If external dependencies (like JSBin) for older snapshots stop working, we're unlikely to go back to all previous snapshots and fix them. We removed JSBins in v1.11.0 mostly in preparation for Ember 2.0 applications which cannot be written as a series of globals in one JS file (so cannot work with JSBin)

PR #1944, opened 2015-01-16
Fix broken jsbin examples in testing guide - Issue #1881

lonelyghost:
Fixed the broken jsbins in the testing guide - they were referencing an older version of handlebars.


fivetanley:
Thanks!

bjornstar:
When will this be live? It's incredibly aggravating to try to use your framework when the official website has broken examples everywhere.

PR #1953, opened 2015-01-25
Correct path to ember-qunit

artfuldodger:
Instead of relying on redirection from GitHub, this change links directly to the repo's current home.


rwjblue:
Thank you!

PR #1954, opened 2015-01-25
Add space before opening curly braces

artfuldodger:
Sometimes those opening curly braces just need more space to breathe.

More importantly, this is more consistent with code elsewhere in the guides.


rwjblue:
Thank you!

PR #1955, opened 2015-01-25
added a short guide on how to use services with ember-cli by subclassing Ember.Service

codepreneur:
This feature is not documented anywhere on Ember website/documentation so I think it would be nice addition for people looking for alternatives to application.inject()

When Ember 1.10 comes out, I think services should get their own section.


locks:
Given the `ember-cli` focus this PR should go towards the effort to rewrite the documentation for module-format/ember-cli.

trek:

PR #1962, opened 2015-01-28
add Toronto Hack Nights and Riverside lat/lng

lzcabrera:
@embersherpa 
I just added the Toronto Hack Nights by specifying the address and a unique name for the location :)

### DIFF
--- a/data/meetups.yml
+++ b/data/meetups.yml
@@ -6,7 +6,7 @@ google_map:
     provider:
       streetViewControl: false
       panControl: false
-      zoomControlOptions:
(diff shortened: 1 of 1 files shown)


wifelette:
I don't think the Riverside meetup was ever merged in, for other reasons, so this won't merge cleanly anymore. 

@lzcabrera If you want to fix so it's just the Toronto Hack Night group (assuming it's a separate group?) we can get this merged :)

wifelette:
Closing for now, can be re-opened if someone wants to pick this back up.

PR #1964, opened 2015-01-28
Update MHElabs link due to it being 404.

jayphelps:
Updated this PR based on the discussion below.

~~Not sure if @mhelabs is still around and active? Maybe @vineet-sinha can chime in? Either way, it's currently a a dead link and being shown on the homepage.  ¯\_(ツ)_/¯~~

image

### DIFF
--- a/data/users.yml
+++ b/data/users.yml
@@ -15,7 +15,7 @@
 # order in which people made their requests. Be respectful to those who
 # came before you and add your name to the *bottom*.
 
-- url: http://www.mhelabs.com/
+- url: http://www.mheducation.com/
   image: mhelabs.png
 - url: https://www.yapp.us/
   image: yapp.png


rwjblue:
MHE Labs is no longer a thing :sob:.  I'll reach out to some of the MHE folks that I am working with to see if they have an alternate link.

habdelra:
You can use this link: http://www.mheducation.com (and yeah that's Will and I on the front page :-)

lukemelia:
Nice Ember Out East shirt!

habdelra:
@lukemelia crazy how that worked out! I think will was wearing with wicked ember shirt too (must have been a meetup that day)

vineet-sinha:
I agree with @habdelra - we should change it to a link to: http://www.mheducation.com

jayphelps:
Updated PR to http://www.mheducation.com/, should be good to merge.

rwjblue:
Thank you!

PR #1966, opened 2015-01-29
Deprecate store.find(type) and store.find(type, query)

wecc:
Update guides to reflect deprecation of store.find(type) and store.find(type, query)

_Don't merge until https://github.com/emberjs/data/pull/2728 has been merged._


trek:

PR #1967, opened 2015-01-29
Edit sample code in finding-records.md

aruseni:
Remove quotes from the query string in an example of a URL that’s requested with store.find.

https://github.com/emberjs/data/issues/2738


teddyzeenny:
Thanks!

PR #1968, opened 2015-01-29
Updated compressed images on top of home

mateuspv:
I always see this images with a little of delay...so I manually compress them.

Note: This not preserve some Photoshop/PNG layer


rwjblue:
Cool, thanks!

PR #1972, opened 2015-01-30
Fix opening codeblock fence

bradfol:
The incorrect opening fence for the code block is causing markdown to render README incorrectly.


rwjblue:
Thanks!

PR #1979, opened 2015-02-04
Fix incorrect docs and add info about coalesceFindRequests

wecc:
Fixes emberjs/data#2745


trek:

PR #1980, opened 2015-02-05
adding Roomle to Ember users

ddprrt:
Mail is out. The new web client for Roomle is based on Ember: http://www.roomle.com/app


wifelette:
Hi @ddprrt —

Can you confirm that this is your company? Basically we want to be sure you're authorized to add their logo. We got your email, but it made me wonder if it was another company of yours, or a friend you were just wanting to add to the site :p Can you clarify?

cc @kmiller915

ddprrt:
I can confirm that. There are two companies, Netural and Roomle, but with about 50% of the same people behind them. I was just asked by my co-workers to add Roomle, because I already added Netural once.

PR #1983, opened 2015-02-07
Getting started remove todo fix

d3chapma:
It was a little confusing to me as I worked through the tutorial when there was code to add the removeTodo method in the step about accepting edits. I moved that code to be added in the Deleting a Model step where it is discussed and makes much more sense IMO.

### DIFF
--- a/source/guides/getting-started/accepting-edits.md
+++ b/source/guides/getting-started/accepting-edits.md
@@ -60,11 +60,6 @@ actions: {
     } else {
       this.get('model').save();
     }
-  },
-  removeTodo: function () {
-    var todo = this.get('model');
-    todo.deleteRecord();
-    todo.save();
   }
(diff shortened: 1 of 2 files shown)


aitherios:
The action `acceptChanges` in `accepting-edits.md` should also be altered as it calls `removeTodo`

Instead of
```javascript
acceptChanges: function() {
    this.set('isEditing', false);

    if (Ember.isEmpty(this.get('model.title'))) {
      this.send('removeTodo');
    } else {
      this.get('model').save();
    }
  }
```

Should be something like
```javascript
acceptChanges: function() {
    this.set('isEditing', false);

    if (!Ember.isEmpty(this.get('model.title'))) {
      this.get('model').save();
    }
  }
```

trek:

PR #1986, opened 2015-02-08
Add example of Query params hbs for context

StillwellMedia:
While this may seem obvious to those familiar with Ember & related concepts, it's helpful to include all steps for new learners.


trek:

PR #1990, opened 2015-02-09
Added example to load model in route when using findQuery to load data

anilmaurya:


trek:

PR #1992, opened 2015-02-10
Wrong parameter used in filter-example

hussfelt:
I'ts currently using the item to check the count, should be using index


wecc:
I think that since the example contains an array of integers, checking `item` is intended. The expected result is defined as `// returns [1,2,3]` (all items with a value less than 4) after your change the returned value is `[1,2,3,4]` (all items where index is less than 4).

hussfelt:
@wecc you're right. Tried enhancing the example, but figured the current one is probably one of the better. Closing.

PR #1999, opened 2015-02-10
Update accepting-edits.md

stebru:
Not sure if this is specific to Ember CLI?


mixonic:
`edit-todo` here is a component, as mentioned in the surrounding text.

stebru:
Hmm, but it was registered as a helper in `js/views/edit_todo_view.js`:
```
Ember.Handlebars.helper('edit-todo', Todos.EditTodoView);
```

It didn't work for me: https://github.com/stebru/hello-ember/blob/master/app/templates/todos/index.hbs
Getting error message: "Uncaught Error: Assertion Failed: A helper named 'edit-todo' could not be found" when using edit-todo without view and quotes. Reproducable here: https://github.com/stebru/hello-ember

But maybe I screwed up something else?

PR #2005, opened 2015-02-12
new user claroapps store

abFx:
Ember new user claro apps store logo


kmiller915:
Email received. cc @wifelette

PR #2007, opened 2015-02-12
Add Blimp to the users page

gcollazo:


kmiller915:
Email received. cc @wifelette

PR #2013, opened 2015-02-14
add Rake task to only generate Ember Data docs

fivetanley:
This is nice when you only want to build/publish ember data docs.

Because Middleman will generate docs for ember with ember data,
this often leads to accidentally changing the ember docs when updating
ember data.


bmac:
+1

PR #2020, opened 2015-02-16
Fixing test jsbins

### LINKED ISSUE #1988: API doc links to source on github are broken
The links to github source from the API docs are broken. 

They look something like 
`https://github.com/emberjs/ember.js/tree//packages/ember-views/lib/main.js#L53`

I believe the cause is the missing `sha` here: https://github.com/emberjs/website/blob/master/data/api.yml#L7

ryanmurakami:
Adding ember-template-compiler.js to some of the jsbins. Upgrading the ember-qunit dependency on all of them.

### DIFF
--- a/data/users.yml
+++ b/data/users.yml
@@ -349,7 +349,7 @@
   image: roomle.png
 - url: http://www.claroapps.com
   image: claroapps.png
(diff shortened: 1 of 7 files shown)


trek:

PR #2024, opened 2015-02-20
WIP Rework Ember Data Guides

bmac:
This is an attempt to rework the Ember Data guides to introduce some of the concepts that the `RESTAdapter `uses along side the basic API. Currently the guides try to keep these 2 concepts separate because it is possible to use Ember Data with other Adapters/Serializers. However, this leads to some confusion for new developers who attempt to use the examples shown in the guide and don't understand why it is failing when their back returns a different format from what the default `RESTAdapter` expects. 

In this pr I'm attempting to follow the pattern of introducing an API. 

screen shot 2015-02-20 at 9 56 58 am

Showing the expected request made my the RESTAdapter and the response from the server. 
screen shot 2015-02-20 at 9 57 04 am

Then introducing the basic hooks to customize the request and how ED extracts the response.

screen shot 2015-02-20 at 9 57 10 am
screen shot 2015-02-20 at 9 57 14 am
 

I'm also tying to reorganize the ordering of the guides to better match how I've seen new comers feel out the features of Ember Data. This mainly results in the finding records guide coming first while the creating and deleting guide is moved till after user has learned how to work with and save existing records. I would also like to remove the "Connecting to an HTTP Server" and "The Rest Adapter" guides because their concepts will be introduced throughout and recapped in "Customizing Adapters". "Using Fixtures" is also removed since it tends make things harder for new users who develop against the FixtureAdapter then attempt to transition to a real backend.

screen shot 2015-02-20 at 9 50 07 am

As always feedback is greatly appreciated.


bmac:
cc @igorT @tomdale @wycats

wycats:
@bmac this is great work!

Just a quick word of warning: As JSON API stabilizes, I think it's going to take a more and more central role in the documentation for Ember Data. For people who can't use JSON API directly, the easiest way to write an adapter will likely be "write a JSON API server in the client".

But none of that really affects your cleanup, so carry on!

bmac:
Note for me to move references to the "the normalized object" from the find guide into the customizing a serializer guide.

locks:
@bmac FYI https://github.com/emberjs/guides/pull/8

trek:

PR #2026, opened 2015-02-20
Update computed-properties-and-aggregate-data.md

fivetwentysix:
Is this the more Ember approach?


mixonic:
The paragraph of text around this specifically talks about `@each`.

PR #2032, opened 2015-02-25
Fixing input event sentence to match current syntax

jasonmit:
Example: http://jsbin.com/variwevuje/2/edit

`on` as an argument to `{{input}}` is no longer valid for most of the events, which the docs suggest to use.


jasonmit:
Interesting, it appears to be working.  Likely an issue with my app where I can't use on.. I'll reopen if I can isolate it.

trek:

PR #2033, opened 2015-02-25
Example of using escapeExpression needs to reference Ember.Handlebars

xmeltrut:


trek:

PR #2035, opened 2015-02-26
fix missing comma in javascript snippet for toggle-all-todos.md

finnigantime:
* Same fix as in #2034 but for the next chapter


finnigantime:
Would have included this in my previous PR but I just saw it now as I'm going through the guide serially. :)

PR #2039, opened 2015-02-28
Add CheckMate to users page.

ascot21:
We've been using Ember in our product for several months now.


kmiller915:
Email received. cc @wifelette

PR #2041, opened 2015-02-28
Update displaying-the-number-of-incomplete-todos.md

diokey:
When i was going through the getting started tutorial, i run into an error on this line. I had to to add 

return this.model.filterBy('isCompleted', false).get('length');

to make it work.


fivetanley:
I think this should be this.get('model') instead of this.model?

diokey:
I'm not really sure. I'm new to Ember. Why? I tried this.model and it works as well.

diokey:
You are right. I just changed it

PR #2042, opened 2015-03-01
Correct example in Handlebars Helpers docs

ATMartin:
The existing example was confusing for me as it's indeterminate whether "aPhotoId" is a static value or a model reference. I modified the example to be more explicit about setting a particular value in the dynamic portion of the href.

I know there's an effort to make the docs more accessible to new developers - I hope this helps! Thanks!

### DIFF
--- a/data/api.yml
+++ b/data/api.yml
@@ -16107,8 +16107,8 @@ classitems:
     ```
 
     ```handlebars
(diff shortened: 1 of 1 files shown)


trek:
The api.yml file is the output of a yuidoc generation process. Check https://github.com/emberjs/website#api-documentation for some more information about making PRs against API docs

PR #2046, opened 2015-03-03
Update guides link on home page.

mehulkar:
Taking too long to clone the repo on conference wifi, so i'm doing this inline without testing :open_mouth:


rwjblue:
Thanks for this! We are likely going to hold off on updating the emberjs.com until we have all of the guide URL's redirecting automatically.

mehulkar:
I figured it was by design. Should I close this?

wifelette:
Seems like yes, so closing now :p

PR #2053, opened 2015-03-06
fix Handlebars namespace in helper example

nikolas:
Use `Ember.Handlebars` instead of `Handlebars`, for consistency with the rest of the code example.


trek:

PR #2058, opened 2015-03-08
Core Team updates

mixonic:
Add @ef4 and @mixonic to the core team page. :tada:


rwjblue:
Still need a @mmun, but this is good for now...

PR #2061, opened 2015-03-10
Example of rejecting a promise along with error route, controller and template

ianwdunlop:
Examples of what a promise rejection can look like. Also basic examples of what to do with it in the error route, controller and template.


trek:

PR #2066, opened 2015-03-13
[Guides] Clarify how to loop through sorted controller content.

whastings:
I found that looping through `{{#each thing in model}}` does not respect the sorting specified by `sortProperties`. So I think adding this clarification will help users who are confused when they are in this circumstance.

### DIFF
--- a/source/guides/controllers/representing-multiple-models-with-arraycontroller.md
+++ b/source/guides/controllers/representing-multiple-models-with-arraycontroller.md
@@ -67,6 +67,16 @@ App.SongsController = Ember.ArrayController.extend({
 });
 ```
 
+To display content in sorted order, loop through `arrangedContent` instead of `model`.
+
+```handlebars
+<ul>
+  {{#each song in arrangedContent}}
+    <li>{{song.name}} by {{song.artist}}</li>
+  {{/each}}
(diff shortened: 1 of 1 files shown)


trek:
Thanks for The PR! Sadly, directly accessing `arrangedContent` is a template is not something we'd recommended. 

http://stackoverflow.com/questions/12998138/ember-js-and-arrangedcontent has some good advice around this question. The guide at http://www.matchingnotes.com/ember-array-proxy/ might also be helpful.

For Ember 2.0, there will be no proxies. You'll either put the model data into the order you want before rendering or leave this up the component that is doing the rendering.

whastings:
Thanks for the clarification. But I still don't see: what should I refer to in my template for `something` when I want to `{{#each thing in something}}` and I want `something` to be my model array sorted by `sortProperties`?

mmun:
`{{#each thing in controller}}` should do it.

whastings:
Thanks! Would it be worth it to resubmit my PR with that syntax for the docs?

trek:
We're migrating currently migrating the guides to current best practices in preparation for 2.0. I think the existing 1.x guides are just going to be as-is for the future.

PR #2068, opened 2015-03-16
Can not inject into helpers.

rwjblue:


rwjblue:
Updated upstream in: https://github.com/emberjs/guides/pull/59

PR #2072, opened 2015-03-18
Add example using model to 1.9 deprecations around context-switching

jdhines:
For newer folk, the use of the word 'model' is helpful, especially for those that are just coming into Ember by following a tutorial.


trek:
We're versioning docs at https://github.com/emberjs/guides so this should really just update that guide to use current best practice.

PR #2075, opened 2015-03-19
Redirects guides to versioned guides

trek:


distilledhype:
<3

joostdevries:
@trek wouldn't a redirect to `/latest` make more sense and then have `guides.emberjs.com/latest/some-page` resolve to whatever is the latest version of that page? Or is the plan to update the erb after every release?

trek:
@joostdevries less sense, I'd imagine. 

Where would if we redirect http://emberjs.com/guides/controllers/ to in Ember 2.0?

joostdevries:
@trek good point.

wagenet:
Seems plausible. What about people linking with bookmarks? Should we add a JS redirect in as well like I did with middleman-alias?

PR #2076, opened 2015-03-20
Move deprecations out of guides

wagenet:
*DO NOT MERGE:* We need to redirect the old deprecations page.

Due to the way we're versioning guides, having deprecations in the guides
no longer makes sense. Since deprecations are major version specific, I've
started a practice of splitting them up by major version.

Right now this page isn't linked to in any navigation, so we may want to
think about this.

See also https://github.com/emberjs/guides/pull/70.


trek:
Looks good to me. I can update https://github.com/emberjs/website/pull/2075#issuecomment-83734612 to point the deprecations guide to this page. @rwjblue any comments?

trek:
#2705 entirely removes the deprecations file. To get internal redirect working, we need to add yaml front matter to `source/deprecations/index.html.md`: https://github.com/Octo-Labs/middleman-alias

rwjblue:
I'm happy with this change in general, once @trek's concerns are addressed.

wagenet:
@trek does middleman-alias maintain the bookmark hash when redirecting?

wagenet:
Looks like middleman-alias does not handle the hash. I made a fork that does. Hopefully, it's a legit solution.

PR #2080, opened 2015-03-21
Add Ann Arbor, Michigan meetup group

tzellman:


trek:
<3. I should stop by when I come visit next. So should @searls

tzellman:
Definitely! It's a great group with lots of excitement. When you're in town next we should coordinate something.

trek:
When/Where do you usually meet?

tzellman:
Well, we are getting things rolling again after a little hiatus. We had two
meetups early last year at the Digital Roots office in Northville, hoping
it would make for a good central location between AA/Detroit/Troy. Now that
the Ember community seems to be heating up a bit more there is higher
interest, and focusing on Ann Arbor makes the most sense for now - although
anywhere in Southeast Michigan would be fine as well.

We just had our first meetup of 2015 on Thursday at the Atomic Object Ann
Arbor office. It was awesome. There was some hardcore nerding out on Ember,
and it was fun.

I've had a few more companies reach out as well, so we may mix up the
location. We're thinking third Thursday of the month, however I think
everyone is pretty flexible.

Having a core team member at a meetup would certainly be amazing.

searls:
+1 that sounds great. @kpfefferle and others are currently figuring out the logistics of setting up a  Columbus group as well :-D

trek:
@tzellman reach out to http://workantile.com/ and tell them Trek sent ya.

twbrandt:
@tzellman and I have already spoken. I would be happy to host the Ann Arbor meetups at Workantile, but unfortunately Thursday evenings are bad for me. Any other day works.

PR #2084, opened 2015-03-23
Add Sellf logo to users page

r4m:
Add logo and website specs


r4m:
our ember.js webapp can be found here http://www.sellf.io

PR #2089, opened 2015-03-24
Add shortcut to the guides at bottom of home page

bengarnett:
help new users take the new step in learning ember.js


wifelette:
Heya—

I'm going to close this for now because we're in the middle of rethinking much of the content on the homepage. We agree that things need to be easier :stuck_out_tongue: It's gotten a tad stale and also cluttered (hence not adding anything new for right now).

PR #2094, opened 2015-03-25
Added link to the API documentation for the model hook

perlun:
I'm not sure if this is the absolutely best place to put it, but I definitely feel that we should have the link in here somewhere. We already have links for `beforeModel` and `afterModel`.


trek:

PR #2099, opened 2015-03-27
Add deprecation entry for using the same function as getter and setter

cibernox:


cibernox:
@rwjblue I'll update the deprecation warning to point to this entry

source/deprecations/v1.x.html.md
```diff
@@ -387,3 +387,37 @@ After:
 
 ```handlebars
 {{#with foo as |bar|}}
+```
+
+#### Deprecate using the same function as getter and setter in Computed Properties
+
+Now that we have the new computed property syntax ([RFC](https://github.com/stefanpenner/rfcs/blob/read-only-cp/0000-read-only-cp-syntax.md))
```
rwjblue: Final location is: https://github.com/emberjs/rfcs/blob/master/active/0001-improved-cp-syntax.md

rwjblue:
One tiny nit-pick about the link to the feature, but other than that it looks good
.

cibernox:
Good catch. Corrected.

PR #2106, opened 2015-03-30
Specify the component helper example in the Ember 1.11.0 blog

jacekkolasa:
the 'either' word would be very helpful for me when I tested this feature, so I wanted to underline that this is an alternative and only 1 colorComponent is rendered.


rwjblue:
Seems good. @mixonic - r?

mixonic:
I'll take it. Thanks @jacekkolasa!

PR #2112, opened 2015-04-01
Updated the guide url to the latest.

### LINKED ISSUE #2111: Click guides link on Community page takes you to the non-latest guides.

bradcypert:
Resolves #2111

In the future, might not be a bad idea to handle /guides/latest or something similar, but this'll solve the problem for now.

### DIFF
--- a/source/layouts/guide.erb
+++ b/source/layouts/guide.erb
@@ -2,7 +2,7 @@
 <html lang="en">
   <%
     @path = current_page.url.gsub(/^\/guides/, '')
-    @new_url = "http://guides.emberjs.com/v1.10.0#{@path}"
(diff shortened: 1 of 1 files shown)


rwjblue:
I believe that the intent was that all previous bookmarks to these URLs will always resolve to the 1.10 guides. 

@trek - can you confirm?

trek:
Confirm. All guides before the revision can be considered `===` to the guides at revision v1.10.0. All bookmarks should redirect to that content. We are not tied to the particular guide structure per revision moving forward.

In fact, we've already removed several guides, so a redirect to 1.11.0 would break old bookmarks to them.

See also: https://github.com/emberjs/website/pull/2075#issuecomment-83729207 for reference.

PR #2116, opened 2015-04-02
[FEATURE] updated about ember examples to 1.11.0

kiwiupover:
So long `bind-attr`

Also updated each helpers to use block syntax.

Also update ember-source to 1.11.0.1. There was an issue with 1.11.0 

Cheers


rwjblue:
LGTM

rwjblue:
Thanks!

PR #2118, opened 2015-04-02
[FEATURE] removing bind-attr from the build ember app

kiwiupover:
RIP `bind-attr` 

Also updated to block params in each helpers

@rwjblue I don't really like the empty string for the truth case.
```js
<div class="left {{if project.beta1Completed '' 'future-image'}}">
```
But this is not any better.

```js 
<div class="left {{unless project.beta1Completed 'future-image'}}">
```

Where is `beta1Completed` coming from?

Cheers

### DIFF
--- a/source/javascripts/app/builds/templates/_files_table.js.hbs
+++ b/source/javascripts/app/builds/templates/_files_table.js.hbs
@@ -2,20 +2,20 @@
 {{#if project.isEmberBeta}}
   <div class="release-progress clear">
     <div class="left">
-      <div class="small-release-image" {{bind-attr title=project.lastStableVersion}}></div>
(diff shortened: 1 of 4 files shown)


rwjblue:
Coming from https://github.com/emberjs/website/blob/master/source/javascripts/app/builds/app.js.

I basically hate that we have this app vendored in this repo.  I would much rather see it live in a different repo as an ember-cli app.  Then we can make things much nicer to work with.  Mostly need to find someone else to do all the work (cause I'm lazy)...

eccegordo:
@rwjblue convert it to a test case for `ember-cli-migrator` and then it will always be kept up to date. :-)

PR #2121, opened 2015-04-04
Add sparehanger to list of ember users

selbyk:


kmiller915:
Email received. cc @wifelette

PR #2130, opened 2015-04-10
add AISLER to list of ember users.

Libermentix:


kmiller915:
Email received. cc @wifelette

wifelette:
The logo is incorrectly sized. Needs to be fixed, and then the PR rebased, most likely.

Libermentix:
what size do we require here? 
Is it sufficient whitespace?

wifelette:
210 x 103 is the space the logo will occupy, so something less than that—ideally you add in the whitespace around it so you're controlling how much of the space your logo fills and it looks the way you want it to :) The previous submission ended up super stretched.

I'm about to commit a change to the file so this info is now provided up top with the rest of the instructions.

PR #2131, opened 2015-04-10
fix controller deprecation warnings for builds app

koriroys:


acorncom:
Seems to have been fixed in #2160

koriroys:
yay open source

PR #2132, opened 2015-04-11
Add Ember CLI link to Navigation bar

changeweb:


wifelette:
Closing—_for now_. All the other links in the main nav stay on the property, so we should't add something that takes you to another domain.

We're working on some big redesign stuff (not far enough along yet to be posted, but just getting started!) that will do a much better job than the current site of making CLI obvious and accessible.

PR #2133, opened 2015-04-12
Fix misspelling “sytnax” in deprecations

roryokane:


rwjblue:
Can you also add a line to preserve the old invalid anchor?

Something like (but with the current value):

```
<a id="toc_warning-when-binding-style-attributes"></a>
```

We link directly to a number of these deprecations from Ember, and we need to make sure that those links (which have already been distributed and can't be updated) still work.

roryokane:
Sure; it’s done.

rwjblue:
Thank you!!

PR #2140, opened 2015-04-15
Add SAKURA Internet to ember users

esukei:


kmiller915:
Email received. cc @wifelette

PR #2141, opened 2015-04-17
Corrected incorrect yielded block param in example

jayphelps:
The block params introduction example incorrectly uses `name` in the template instead of `fullName`.

http://emberjs.com/blog/2015/02/07/ember-1-10-0-released.html#toc_block-params


jayphelps:
"Corrected incorrect". heh....it's late.

rwjblue:
Technically the naming used is not incorrect.  What the consumer chooses to name a given block param in their template is not related to what the component creator called the same param (they are ordered arguments not named args or hash args).

Obviously, that seems somewhat pedantic (and isn't mentioned in the blog post anyways).

PR #2142, opened 2015-04-18
Add Ludu as an Ember company

t4t5:
Email has been sent to listing@emberjs.com. We're big fans of Ember!


kmiller915:
Email received. cc @wifelette

PR #2146, opened 2015-04-21
Fix blog title formatting when the line wraps.

matthewlehner:
This is just a small tweak for the blog titles on actual blog post pages. The [IE8 Support Update](http://emberjs.com/blog/2015/04/20/ie8-support-update.html) post, in particular.

Here's the before:
screenshot 2015-04-21 08 56 04

And with this commit:
screenshot 2015-04-21 09 00 56

### DIFF
--- a/source/stylesheets/blog.css.scss
+++ b/source/stylesheets/blog.css.scss
@@ -1,6 +1,5 @@
 article.blog-post {
   h1 {
-    height: 24px;
     padding-left: 25px;
   }
 


trek:
What does this look like on titles that don't wrap?

matthewlehner:
Ah crap. Why didn't I check that? It cuts off the bottom of the ribbon. I'll fix this and send another PR.

PR #2147, opened 2015-04-21
Fix ribbon on blog post titles that don't wrap.

matthewlehner:
This fixes a regression on single line post titles caused by #2146


matthewlehner:
For reference – here's what a single line post title will look like:

screenshot 2015-04-21 09 30 11

And a wrapping title:
screenshot 2015-04-21 09 31 00

knownasilya:
Just saw the issue, and was about to go see if it needed fixing; looks like you beat me to it.

PR #2149, opened 2015-04-22
Add aisler to list of ember users.

Libermentix:
new pull request as discussed in #2130 with @wifelette


wifelette:
This needs to be rebased now, but also, see my previous comment about whitespace. Did you stage this with the logo this size? I see a tiny amount of whitespace, so it might be good, but just making sure you took a look at it.

Libermentix:
@wifelette Yes I did. It should be fine. Thanks for your advice and sorry for the hassle ;-)

wifelette:
Fantastic :) And no hassle... still needs to be rebased so we can merge though :p

Libermentix:
@wifelette I think i managed it now ;-) Had to dig into merging/rebasing before.

wifelette:
Yay! All good to go now :)

PR #2150, opened 2015-04-22
Add Ember Columbus Meetup

rickygi:


data/meetups.yml
```diff
@@ -371,6 +371,17 @@ locations:
     organizers:
     - organizer: Tom Zellman
       profileImage: http://photos2.meetupstatic.com/photos/member/d/a/c/c/member_91316012.jpeg
+  - location: Columbus, Ohio
+    url: http://www.meetup.com/ember-columbus/
+    lat: 39.9833
+    lng: 82.9833
```
kpfefferle: I think this is supposed to be a negative value for longitude. Otherwise you end up on the map in China :)

data/meetups.yml
```diff
@@ -371,6 +371,17 @@ locations:
     organizers:
     - organizer: Tom Zellman
       profileImage: http://photos2.meetupstatic.com/photos/member/d/a/c/c/member_91316012.jpeg
+  - location: Columbus, Ohio
+    url: http://www.meetup.com/ember-columbus/
```
kpfefferle: Should be updated to http://embercolumbus.com (which currently redirects to Meetup) so that we don't have to change it later should we end up building out more content there.

kpfefferle:
The meeting space is at the following coordinates:
```
lat: 40.098794
long: -83.105245

PR #2158, opened 2015-05-01
eflex systems removed tag-line from image base

ericop:
My apologize for excessive pull requests guys. Guy who made the call on design as out of the office yesterday, but I think we've got it straight.


wifelette:
Can you shrink this image? Expected size for the logos is filling a space 210 x 103 (so you want to be *smaller* than that so you have some whitespace).

ericop:
okay @wifelette does that look better?

PR #2159, opened 2015-05-01
Add Vestorly to ember-users

greggilbert:


kmiller915:
Email received. cc @wifelette

PR #2161, opened 2015-05-04
Update beta build to 1.12.0-beta.3

josemarluedke:
The dates were just a guess, I'm not sure if that's the real schedule. Could someone confirm if that's correct or not?


rwjblue:
Thank you!

PR #2169, opened 2015-05-11
add isle of code to list of ember users

AlexBlom:


kmiller915:
Email received. cc @wifelette

PR #2173, opened 2015-05-14
Ember 1.12 blog post, fix instance initializers example

pixelhandler:
`export default` should only be used once, the example uses it twice. I tried that and an an error.


igorT:
Thanks!

PR #2175, opened 2015-05-15
Fix getter/setter syntax

dustinfarris:


rlivsey:
The syntax is valid, it's using the [new ES6 object literal syntax](http://babeljs.io/docs/learn-es6/#enhanced-object-literals).

dustinfarris:
ah, ok

PR #2179, opened 2015-05-19
Stick footer to the bottom of the document

dbbk:
Improves the appearance of the site's layout when viewed with tall monitors, or when sections such as Builds are in the process of loading.


wifelette:
@dbbk This is super old, so hopefully you'll still pay attention <3

What browsers are you trying to support here that don't support the standard version of the syntax?

wifelette:
Closing this for now, since it seems to have stalled. Just pop back around if you want to take another stab at it :)

PR #2185, opened 2015-05-22
Add Libro to ember.js users

jdurand:


kmiller915:
Email received. cc @wifelette

PR #2203, opened 2015-06-12
Note on array/reduce computed being deprecated

cibernox:
The last paragraph is actually not true until https://github.com/emberjs/ember.js/pull/10582 gets merged


rwjblue:
@cibernox - Can you remove the last section for now?  I'd like to get this landed and published now, then have a second PR to add the bit about those macros being replaced with plain array versions.

cibernox:
Right now

cibernox:
done

PR #2210, opened 2015-06-13
Fix an incorrect closing tag for `{{#reverse-name}}`

martypenner:


rwjblue:
Awesome, thank you!

martypenner:
Thank you for putting so much hard work in!

PR #2217, opened 2015-06-19
Syntax error in example.

Criten:


igorT:
This is not a syntax error, it uses the new ES6 method syntax https://github.com/lukehoban/es6features#enhanced-object-literals

PR #2219, opened 2015-06-23
Add more explicit referencing of deprecations guide

typeoneerror:
The deprecations guide is chock-full of useful information for those working with builds other than stable. Let's reference it on that page.

Also includes some CSS cleanup in the "builds" application. Looked like there were some leftovers.


mixonic:
This seems great, thanks @typeoneerror

PR #2220, opened 2015-06-23
Do not share `content` on prototype.

rwjblue:


rwjblue:
@bantic - Can you update the JSBin?

source/deprecations/v1.x.html.md
```diff
@@ -704,9 +704,17 @@ dropdown options, the following code could be used:
...
 export default Ember.Component.extend({
-  content: [],
+  content: null,
   selectedValue: null,
 
+  didInitAttrs(attrs) {
+    var content = this.get('content');
+    
+    if (!content) {
+      this.set('content', []);
```
stefanpenner: just put this in `init`

rwjblue: `init` is not guaranteed to have attributes. It does for 1.13 but for angle bracket components it will not, might as well do it right...

bantic:
@rwjblue updated the jsbin (and removed deprecations for  `key="@guid"` in `{{each}}` and generated arraycontroller).

I'll follow this issue while it's open and keep it up to date if the discussion changes.

source/deprecations/v1.x.html.md
stefanpenner: we are just setting the initial value, `init` is the correct place for that.

bantic: I changed the [jsbin](http://jsbin.com/fotuqa/edit?js,output), but I'm not sure if it still exhibits the same behaviors.

The blog post announcing 1.13 seems to imply that `didInitAttrs` is the place to do this setup. I read that line "passed attrs are guaranteed to be present" as meaning that passed attrs are not guaranteed to be present in init, which implies that if the `content` key on the component is `null` and the passed attr "`content`" is also null, then it might end up setting the final value to null (if `didInitAttrs` happens after `init` is done, blowing away the set `[]` with the passed-in `null`). Here's the part of the blog post I'm referring to: http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_component-lifecycle-hooks

I am able to reproduce this in [this jsbin](http://jsbin.com/dowola/edit?html,js,output), forked from the other one. It has a controller prop `nullContent` that it passes in to "no passed selection" component section. And instead of init-ing content to `[]` in `init`, in this jsbin it inits it to an initial value (`{title:'x', id:4}`), and logs that it is doing so. If you look at the console logs you can see that it sees that `content` is not set in `init` so it sets it to the default value, and then later the passed-in `nullContent` (value: `null`) overrides it, so the value the component's template sees for `content` is null instead of what we set in `init`: http://take.ms/ebVZE

stefanpenner: that is unfortunate, as it means will have to do the following:

```
export class extend Component {
  init(attributes) {
    attributes.content = attributes.content || [];
    super(attributes);
  },

  didInitAttrs(attrs) {
     var content = this.get('content');
    
    if (!content) {
      this.set('content', []);
  }
}
```

stefanpenner: this also then wouldn't work correctly with the proposed ES7 property initializers.

stefanpenner: I reviewed the code, we can likely fix the internals. At a later date. It is unfortunate for now, but i think we can deal with it.

stefanpenner: this hook should also call `super`

mixonic:
Added the `_super` call in a cleanup commit.

rwjblue:
Thank you @mixonic!

PR #2244, opened 2015-07-06
Fix another typo in ED release announcement

dschmidt:
status is a number

Both according to docs http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_handleResponse and to own experience :)


fivetanley:
thanks!

PR #2248, opened 2015-07-08
Added codechef.com to Ember.js users

skbly7:
Some example of places where it has been used :
(if needed for verification purpose)

IDE
http://www.codechef.com/ide

RANKINGS
http://www.codechef.com/rankings/COOK59 (example URL)

and many others... :)


kmiller915:
Email received, cc @wifelette

PR #2249, opened 2015-07-09
link to slack organization

baijum:


mixonic:
@baijum I strongly believe we should coalesce around a single slack organization. The the one you've added here newer, or has it perhaps been superseded by the `ember-community-slackin` org?

baijum:
The ``ember-community-slackin`` is a heroku app, right? I think this is created to make joining easy.

Well, I am new to this. May be someone else can confirm.

kevinkucharczyk:
Yeah, the herokuapp is for requesting invites to the actual ember community slack organization (you receive the slack link in an e-mail). I think it would be good to keep both links and specify what they are (signup form vs actual slack link).

rwjblue:
@mixonic - The slackin link is how to get an invite, the other link is to the actual slack instance once invited.  They both reference the "Ember Community Slack" that we know and love.

rwjblue:
@baijum - Can you tweak it a bit to make it clearer that the slackin link is for joining and the other is for logging in?

JKGisMe:
Before there was only a link to _sign up_ for slack. Took me a few mins to actually find the slack again the first time I tried to go back to slack when I first joined. I think it could be made more clear what what the two links are.

runspired:
@mixonic the links are correct but @rwjblue is also right that they need context.  The heroku app is so you can invite yourself (Slack requires an admin to send the invite, this app handles that for us).

baijum:
@rwjblue @runspired @mixonic I have given the links with context. I just re-based against upstream master, squashed my commits and pushed. Let me know if this is fine.

PR #2250, opened 2015-07-09
Update the Ember Data beta.19 blog post to say Ember Data 1.13 supports IE8

bmac:
screen shot 2015-07-09 at 9 53 07 am


mixonic:
@bmac thanks!

PR #2251, opened 2015-07-13
Show tagged builds for v2.x

pdud:
Fixes https://github.com/emberjs/ember.js/issues/11725

- Problem: tagged v2.x builds were not showing because S3 returns a max
  of 1000 builds alphabetically

- Solution: recursively load additional pages if the response's
  `isTruncated` field is true

One thing to be aware of is that it sets `response` on the S3Bucket to be the the last response. As far as I can tell `response` is no longer used so it could be removed or left in place.

https://github.com/pdud/website/blob/01dea25589cb0bf5f3376a9a1b53b53ba1c21949/source/javascripts/app/builds/app.js#L144


rwjblue:
Tested locally, looks great, thank you!

PR #2253, opened 2015-07-13
Use consistent `type` attributes for JSON-API data

sandstrom:
I think these names must be consistent, i.e. cannot mix singular and plural (which is currently the case).

The spec is agnostic to singular/plural as long as they are consistent, but seem to use plurals so I did likewise.

I guess Ember Data is using either singular or plural to match with models. If it's using singular, we need to switch these.


sandstrom:
Closed in favour of https://github.com/emberjs/website/pull/2254

PR #2254, opened 2015-07-13
Fix minor inflection issue in ED blog post

sandstrom:
`relationships` keys must match `type`


wecc:
As this is supposed to be a has-many relationship (as in `accounts: DS.hasMany('account')`) I think `accounts` is the correct form already.

sandstrom:
Yes, you are right. I re-read the spec now, it's the `type` attribute that must be consistent. My mistake.

Issue #2256, opened 2015-07-15
Missing Deprecations in Deprecations Guide


Gaurav0:
Please add to the deprecations guide here: http://emberjs.com/deprecations/v1.x/.

These are all of the deprecations in 1.13.x that need to be covered:
 
- http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html
  - [x] `Ember.CoreView` (63deb39ca7db138fe019765af79289131f18556f)
  - [x] `Ember.View` (#2195)
  - [x] `Ember.CollectionView` (#2265)
  - [x] `Ember.ContainerView` (#2265)
  - [x] `{{view 'some-helper'}}` (#2195)
  - [x] The `{{view}}` keyword for accessing properties on a view (#2198)
  - [x] The `{{controller}}` keyword for accessing properties on a controller (#2198)
  - [x] Options to the `{{#each` helper that trigger a legacy and poorly performing legacy layer. These options are: `itemView`, `itemViewClass`, `tagName`, `emptyView` and `emptyViewClass` (https://github.com/emberjs/website/pull/2328)
  - [x] The itemController argument for `{{#each`. (https://github.com/emberjs/website/pull/2328)
  - [x] `bind-attr` (#2305)
  - [x] Reading `this.get('template')` to check for a yielded block on components. (#2433)
  - [x] Non-block param `{{with` (e6cd07b8155c4e71c378d796f89df84ad4a9b14a)
  - [x] The view and viewClass params for `{{outlet}}` (#2433)
  - [x] `{{#each foo in bar}}{{/each}}` (e6cd07b8155c4e71c378d796f89df84ad4a9b14a)
  - [x] `Ember.EnumerableUtils` (#2433)
  - [x] Allow deprecated access to `template` in component to determine if a block was provided. (#2433)
  - [x] `Ember.LinkView` (#2197)
  - [x] `Ember.computed.any` (#2313)
  - [x] `Ember.Select` (#2201)
  - [x] `Ember.arrayComputed` (#2203)
  - [x] `Ember.ReduceComputedProperty` (#2433)
  - [x] `Ember.ArrayComputedProperty` (#2433)
  - [x] `Ember.reduceComputed` (#2203)
  - [x] `beforeOberver` (#2246)
  - [x] Modifying a property inside of `didInsertElement` (Reported in https://github.com/emberjs/ember.js/issues/12290) (#2433)
[...]

PR #2258, opened 2015-07-17
Add ArrayController to 1.13 deprecations

pdud:
This starts some of the deprecations mentioned in #2256


mixonic:
@pdud thanks sir :-) If you are interested, we could use a followup PR to Ember itself that adds a url to array controller deprecations linking here.

pdud:
Thanks for the followup PR idea. Added! :)

Issue #2261, opened 2015-07-20
Add to deprecation guide for Ember.ContainerView


martndemus:
`ContainerView` will become deprecated. See emberjs/ember.js#11818 and emberjs/ember.js#11826.
I'd liked to do it myself but I have very little experience with it myself. Therefore I think someone with better experience with `ContainerView` should write it.

The deprecation message will point to the url `http://emberjs.com/deprecations/v1.x/#toc_ember-containerview`

trek:
Closing in favor of #2265

PR #2265, opened 2015-07-25
Add deprecation guide for ContainerView and CollectionView

martndemus:
It's not much but at least it's something. Maybe someone with better writing skills should improve upon this.

Addresses: #2261


wifelette:
Please check this off the list on #2256 once merged :)

PR #2266, opened 2015-07-25
Add deprecation guide for Controller#needs

martndemus:
Again, i'm not a great writer, but at least there is some documentation.

See emberjs/ember.js#11767


rwjblue:
Thank you!

Issue #2264, opened 2015-07-25
It appears that toggling inherited checkbox enables private members in api docs


petey:
Here's a test case for you:

* Navigate to http://emberjs.com/api/classes/Ember.ArrayProxy.html
* Click on the checkbox labeled inherited
   * All methods disappear
* Click on checkbox labeled inherited
   * More methods appear than existed before (includes private methods)

stefanpenner:
relevant HTML: https://github.com/emberjs/website/blob/master/source/_api_options.html.erb
relevant JS: https://github.com/emberjs/website/blob/59837b08c5bbcbe6579094e78c4f0d15d0139a17/source/javascripts/app/api/api.js

chriskrycho:
This comes up because of an overlap between the types. Some elements are *both* `inherited` and `private`. The toggle behavior does not distinguish between them; it just toggles the value for *every* `inherited` member, and that includes some `private` members. (Notably, not all: if you go ahead and click the Private checkbox, you'll see that the non-inherited private members appear.)

Where is the index list implemented?

ATMartin:
I may have a fix for this but I'm juggling an old bad commit upstream of my fix. Will be just a few on the PR for review.

MaximAbramchuck:
@stefanpenner need to close this one.

PR #2267, opened 2015-07-25
Fix unexpected toggling of displayed items in the API panes.

ATMartin:
In response to #2264: Now we're going through each of the pane's api-option inputs and confirming that they're toggled appropriately on each change. This appears to fix the issue of "Show inherited, show private, unshow inherited, wait why's that still there, oh god make them go away."

Hope this helps!

### DIFF
--- a/source/javascripts/app/api/api.js
+++ b/source/javascripts/app/api/api.js
@@ -23,9 +23,13 @@ $(function() {
       ls.setItem('api-options-' + type, this.checked);
   }
 
+  function confirmPaneInputs() {
+    $(this).closest('#api-options').find('input').each(toggleType);
+  }
+
   $('#api-options input').each(initApiOptions);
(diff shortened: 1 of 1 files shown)


chriskrycho:
Testing this locally (since I was also working on a solution), and it has the side effect of toggling the contents of other panes (but not their checkboxes). E.g., check Private on the Index tab, then click the Methods tab: both Inherited and Private methods are shown there as well, but the checkbox for Private is unchecked.

The behavior in #2264 is fixed along the way, though!

ATMartin:
Ahh, good catch, thanks! Will see if I can nudge that out of there.

ATMartin:
I probably could have just `--amend`ed that into the original. This lets the selections persist across panes though. Newish to this, just let me know if I need to squash those changes together.

PR #2268, opened 2015-07-27
Change method/property/event ordering from linewise to columnwise

### LINKED ISSUE #2263: Methods/properties layout order
Currently the docs list properties like this:
```
Apple          Apostrophe
Anger          Bible
Boot           Train
Xylophone      Zebra
```
I find this very difficult to find a result in.
I'd prefer:
```
Apple           Boot
Apostrophe      Train
Anger           Xylophone
Bible           Zebra
```
screen shot 2015-07-24 at 6 33 17 pm

AaronSikes:
Fixes #2263 
Style changes only.
[Browser Support](http://caniuse.com/#feat=multicolumn)
This should work in all major browsers. In <= IE9 I believe this will fall back to a single column twice as tall. I think the increase in comprehensibility for everyone else is worth the tradeoff.
Before:
screen shot 2015-07-26 at 8 05 13 pm

After:
screen shot 2015-07-26 at 8 05 53 pm

### DIFF
--- a/source/stylesheets/site.css.scss
+++ b/source/stylesheets/site.css.scss
@@ -2137,11 +2137,14 @@ a.edit-page {
   ul.index-list {
     margin-left: 1.5em;
     overflow: hidden; /* Clearfix */
+    -moz-column-count: 2;
+    -moz-column-gap: 1em;
+    -webkit-column-count: 2;
+    -webkit-column-gap: 1em;
+    column-count: 2;
+    column-gap: 1em;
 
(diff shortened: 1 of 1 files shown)


rwjblue:
@AaronSikes - can you also update https://github.com/emberjs/api accordingly?

rwjblue:
Thank you!

AaronSikes:
Sweet! When/how often is the site deployed?

Issue #2276, opened 2015-08-03
Don't uppercase API headers.


maxcal:
![image](https://cloud.githubusercontent.com/assets/765647/9037466/ffbaf660-39eb-11e5-8f5e-ef725edbc57e.png)

*Now is it JSONApiSerializer or JSONAPISerializer  or ...*

While the uppercase headers might look nice they are a usability disaster as it makes it difficult to figure out what the components are actually named - which is pretty dang important for an API in a case sensitive language!

Removing the text transform for the API headers is trivial for a big usability boost.
```CSS
body.api h1 {
    text-transform: none;
}
```

![image](https://cloud.githubusercontent.com/assets/765647/9037532/8ddc72ac-39ec-11e5-8176-bcde755cfe86.png)
 
Will try to write a PR time and energy permitting.

trek:
handled by https://github.com/emberjs/website/pull/2277

PR #2277, opened 2015-08-03
No uppercase text transform for API headers

maxcal:
Improves the readability of the API docs by removing the uppercase text transform. See #2276


rwjblue:
Seems good to me.

@mixonic / @trek / @emberjs/guides-managers - Any thoughts/objections?

michaelrkn:
cc @locks

mixonic:
I am +1.

trek:
Thanks! I've always disliked this.

PR #2279, opened 2015-08-07
Fixing links in 2015-06-18-ember-data-1-13-released.md

pjhauser:
Fixing links to http://jsonapi.org/format/, self referencing links formatted like this: [http://jsonapi.org/format/#document-resource-objects]() we're resolving to http://emberjs.com/blog/2015/06/18/ember-data-1-13-released.html


mixonic:
Thanks @pjhauser

PR #2283, opened 2015-08-08
Make Ember Users page accessible by adding alt tags

Robdel12:
Had some time to kill and noticed none of the images on the page had alt tags.

### DIFF
--- a/data/users.yml
+++ b/data/users.yml
@@ -20,406 +20,607 @@
 
 - url: http://www.mheducation.com/
   image: mhelabs.png
+  alt: McGraw-Hill
 - url: https://www.yapp.us/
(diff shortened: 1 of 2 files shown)


rwjblue:
This is extremely awesome, thank you @Robdel12!!!

wifelette:
Oh man, amazing. I was just in there fixing some other things and was kind'v like... "I don't remember doing that, I guess I'm more awesome than I realized!" 

;)

Update: _you're_ awesome, instead. Thanks so much!

Robdel12:
@wifelette heh, this got lost in my Github notifications. No problem! You're still awesomeer! 

Thanks for the kind words y'all :)

PR #2286, opened 2015-08-08
Add Movable Ink. Fix typo

ksin:


wifelette:
Hey! Thanks for the typo fix!

I haven't pulled this down yet to test, but the logo doesn't _look_ like the right dimensions. I'd hate to see it smashed in either direction! 

Can you verify that it is correct as is, or update the image file to fit correctly? Reminder that the space it'll get put into is 210 x 103. For best results, submit an image exactly that size that already includes as much breathing space around the logo as you'd like.

ksin:
You're right! I'll fix the image real quick.

ksin:
Amended the fix over. Thanks. :)

PR #2290, opened 2015-08-10
add coders51

keown:
add coders51 to ember users


wifelette:
Heya—

Can you do a quick rebase? Also, I'd kill the "for the entire galaxy" bit :p Witty, but the info is there for later parsing and reorganizing, and added fluff would make that more difficult.

keown:
done :+1:

PR #2298, opened 2015-08-11
Add deprecation for reversed observer arguments

ktornwall:
Adds missing deprecation description for Ember 1.13.5 - Deprecate reverse argument ordering in ```Ember.observer```.

See #2256 for checklist of all missing deprecations.


rwjblue:
:+1: - Thank you!

PR #2300, opened 2015-08-11
Add Privatt to Ember.js - Users

jimbeaudoin:


kmiller915:
Email received. cc @wifelette

PR #2313, opened 2015-08-14
Added deprecation guide about Ember.computed.any

phkavitha:
My first commit towards fixing #2256.


wycats:
@phkavitha Thanks for submitting this. I look forward to merging it.

A couple of things:

1. We use [American spelling](https://en.wikipedia.org/wiki/Wikipedia:List_of_spelling_variants) for `favor` (`o` rather than `ou`).
2. We generally add a colon to a sentence that is immediately followed by an example (`For example:` rather than `For example,`).
3. The reason we changed from `any` to `or` is to reflect the fact that the computed value is the first successful value (like `||`) rather than a boolean (like `Array.prototype.any`). It might be useful to include that information.

Thanks so much again for the contribution!

wycats:
Also, can you rebase?

phkavitha:
@wycats , I have made the necessary changes. Let me know if any changes are required. Thanks for your help.

rwjblue:
Thank you!

PR #2315, opened 2015-08-14
Add note with the only breaking change in Ember 2.0: Ember.set

cibernox:


cibernox:
@mixonic @stefanpenner Brought by https://github.com/emberjs/ember.js/issues/12095

stefanpenner:
Can you include the rationale. Proper es semantics and sanity with super in cp

cibernox:
@stefanpenner can you remind me the problem with `super`?

stefanpenner:
it was the same as:

```
computed('other', {
  get() { return this.get('other'); },
  set(_, v) { 
    // some code that is a setter
    return this.set('other', v); // returns the wrong thing
    // vs 
    return this._super(); // returns the right thing
   }
})
```

rwjblue:
seems like this needs a rebase

PR #2317, opened 2015-08-15
Update v1.x.html.md

glagola:
A code block must be separated from text above, otherwise it will be rendered incorrectly. [Currently](http://emberjs.com/deprecations/v1.x/#toc_ember-collectionview)


rwjblue:
Thanks!

PR #2319, opened 2015-08-16
add TPP to ember user list

leojh:


kmiller915:
Email received. cc @wifelette

wifelette:
Hi @leojh —please update the Yaml. Use, as in, how are you using Ember, not what does your product do ;) 

Also, can you confirm that your logo is the exact correct size? I'm not on a machine where I can pull it down and check right now.

leojh:
Hi Leah. I meant to say that we are using ember to modernize our app but I will re write the use to be more clear. Also, the image should be the correct size and with a transparent background.

leojh:
Let me know if this new use works for you :)

PR #2334, opened 2015-08-20
Add Kickstarter as an Ember user

praxxis:
Email incoming


kmiller915:
Email received. cc @wifelette

PR #2338, opened 2015-08-20
Add Ember Weekly to community links

owilliams:
The community page used to link to Ember Weekly before the Meetup page redesign, where the community links were removed. It looks like a few have been added back on now and I was hoping Ember Weekly could potentially be linked? The community page used to drive quite a bit of traffic to Ember Weekly in the past. 

Let me know if you think it's appropriate or not. Thanks!

### DIFF
--- a/source/community.html.erb
+++ b/source/community.html.erb
@@ -71,6 +71,9 @@ title: "Community"
   <p>
     For a curated list of Ember learning resources (podcasts, videos, blog posts, books, etc) check out <a href="http://emberwatch.com/">EmberWatch</a>.
   </p>
+  <p>
+    Stay up to date with the latest news and resources with the regular email newsletter, <a href="http://emberweekly.com">Ember Weekly</a>.
+  </p>
 </div>
 
 <div class="community section meetup">


wifelette:
Hi Owain—

In general, we aim to avoid linking to (aka implicitly endorsing) revenue-generating or commercial operations, and specifically the sort of things that other community members might want to and could legitimately compete with. StackOverflow is a bit of an unusual case, I know, but it's not really comparable (in the same way we'd still link to a Wikipedia page, though they do periodic aggressive fundraising). 

We _also_ aim to keep the resources list on a page like this fairly short—essentials, basically. EmberWatch (non-commercial as far as we know) was added as an alternative to building something similar on our own, more of a bandage than anything, and something we're still discussing. This is primarily to avoid needing to _maintain_ this list, but also to avoid aligning to closely with any particular thing and accidentally stifling competitive desires others might have. Best bet for everyone is if there are numerous Ember resources :) 

I can let you know if we think we want to make an exception, but my _instinct_ is that this wouldn't be the sort of thing we want to add there.

PR #2344, opened 2015-08-23
Expose rake task to generate just the ember docs (not data)

EricSchank:
Create rake task to just generate the ember side of the docs.
(There's already one for just the ember-data side.)
This is helpful if you're testing doc changes to only one side of the project.


rwjblue:
Thanks!

Issue #2310, opened 2015-08-13
"Writing Helpers" link goes to v1.10


locks:
"Writing Helpers" link at https://github.com/emberjs/website/blob/master/source/deprecations/v1.x.html.md#L352 is default to `v1.10`.

michaelrkn:
This will be fixed when we have `/current` in the Guides.

PR #2345, opened 2015-08-23
Redirect guides to v2.0.0

josemarluedke:


josemarluedke:
Possibly fixes #2310.

josemarluedke:
ping.

acorncom:
@wifelette I think this would be fine to merge here ...

wifelette:
I think @michaelrkn had some thoughts on this in the last call?

michaelrkn:
What we really need is https://github.com/emberjs/guides/issues/604 so that we don't have to keep updating this, but at least this is better than how it was before. Thanks @josemarluedke!

PR #2349, opened 2015-08-27
Fix Ember.keys Deprecation Example

elwayman02:
`Object.keys` only takes a single parameter, but the example was passing multiple objects for some reason, rather than a single hash with multiple attributes.


rwjblue:
Awesome, thank you!

PR #2371, opened 2015-09-27
Added Global Ember Meetup to meetups page

taras:


wifelette:
@taras Where do the map coordinates for this point? I can't pull down to stage right now, but I plugged it into an online lat/long map and it seemed to think they were invalid coordinates.

taras:
The coordinates are pointing into the middle of North Atlantic ocean. I'm not sure what else to use :)

wifelette:
Let's call the category Online, not Global, and leave it off the map altogether :)

Gotta make sure that omitting that info from the Yaml file doesn't blow anything else up though, and correct if so. Please verify and update PR and we should be good to go!

taras:
I need to give it some "location" which is used as a name. What should I use there?

wifelette:
Ah, you mean because if Online is the category, having the name of the specific meeting be Online is repetitive. Hmmm.

I'm not sure. The idea is that there may be more online meetups in the future. But I'm unsure how you'll each differentiate yourselves. And the listing design is mostly optimized for short-form names.

Ideas?

wifelette:
Closing due to inactivity.

acorncom:
To clarify, we're closing this pull request due to inactivity, not because we wouldn't be happy to take a pull request that addresses concerns, correct?

wifelette:
Mhmmm

PR #2384, opened 2015-10-05
Add info about deprecated function as second argument in deprecate/warn/assert

Kuzirashi:
As requested by @rwjblue in [comment](https://github.com/emberjs/ember.js/pull/12370#issuecomment-145419368) in [pull request](https://github.com/emberjs/ember.js/pull/12370) which added deprecations.


rwjblue:
:heart:!!!

PR #2391, opened 2015-10-19
Adds Practice Fusion to the list of ember-users

MiguelMadero:
Adds Practice Fusion to the list of Ember Users


kaddae:
Email received CC @wifelette

MiguelMadero:
thanks!

PR #2392, opened 2015-10-20
Remove G+ link from home page

acorncom:
The last time the G+ page was updated was in 2013.  Seems best to remove it ;-)


rwjblue:
Ha! Agreed!

PR #2395, opened 2015-10-29
fix App Instance example: use instance initializer

gabrielgrant:


rwjblue:
Great catch, thank you!

gabrielgrant:
Thanks for getting it merged! Hopefully it'll save someone else a wee bit of confusion :)

PR #2397, opened 2015-11-02
Stop using multiple `source` in primary

tricknotes:
The following warning is shown since bundler 1.7.

```
Warning: this Gemfile contains multiple primary sources. Using `source`
more than once without a block is a security risk, and may result in
installing unexpected gems. To resolve this warning, use a block to
indicate which gems should come from the secondary source. To upgrade
this warning to an error, run `bundle config disable_multisource true`.
```

See detail: https://github.com/bundler/bundler/blob/master/CHANGELOG.md#170-2014-08-13

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,4 @@
 source 'https://rubygems.org'
-source 'https://rails-assets.org'
 
 gem "redcarpet"
 gem "activesupport"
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

PR #2403, opened 2015-11-04
add return value for setter

et:
Setters expect the value to be returned, see [docs](http://emberjs.com/api/classes/Ember.ComputedProperty.html#method_set).
I got bit by this and wasted about half a day trying to figure out why property changes weren't being fired.


rwjblue:
Sorry that this caused you so much trouble, thanks for fixing it for others!

et:
No worries! I learned a lot about ember internals so it was not a lost cause :smile:

PR #2405, opened 2015-11-06
Added deprecation for `Ember.SortableMixin`

workmanw:
Documentation isn't my strong suit, happy to make any changes.

Also, I added this deprecation immediately after Ember.ArrayController because they are some what related and it seemed like they should be next to each other.

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -1001,7 +1001,7 @@ In many cases you can move your logic later in the component lifecycle by implem
     this._super(...arguments);
     this.set('myValue', value);
   }
-``` 
+```
 
 #### beforeObserver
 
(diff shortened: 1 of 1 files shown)


source/deprecations/v1.x.html.md
```diff
@@ -1023,6 +1023,38 @@ proxying behavior is being used.
...
+
+Along with `Ember.ArrayController`, `Ember.SortableMixin` will be removed in Ember 2.0.
+
+You can migrate away from using `Ember.SortableMixin` by using `Ember.computed.sort`. Using this example:
+
+```
+import Ember from 'ember';
+
+export default Ember.ArrayController.extend(Ember.SortableMixin, {
+  model: songs,
```
stefanpenner: Putting state, like `model` on the prototype isn't a good idea.

source/deprecations/v1.x.html.md
```diff
@@ -1023,6 +1023,38 @@ proxying behavior is being used.
...
+});
+```
+
+You can transition to using `Ember.computed.sort` like this:
+
+```
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  model: songs,
```
stefanpenner: ^

wifelette:
@workmanw Thanks for taking a stab at this! I got @wycats and @stefanpenner to review, see Stef's feedback. Looking forward to merging once it's resolved!

source/deprecations/v1.x.html.md
workmanw: I was just following the other examples for consistency ... I think they were assuming that `songs` would be populated via the router. I can break with the pattern here though.

Would you like to see a something thinking this:
```
  init: function() {
    this.super();
    this.set('songs', [ ... ]);
  }
```

Or would you rather me just emit the `model` definition all together and provide a footnote explaining that `model` is set by the route?

acorncom:
@workmanw Could you rebase here to resolve conflicts?  We'll see about getting this further attention next week ...

workmanw:
@acorncom Yes. I rebased this.

acorncom:
@workmanw :+1:

acorncom:
@workmanw Per a discussion in the docs meeting today, this is not getting merged as it needs attention from people who know it's history well.  Will keep you updated

workmanw:
¯\_(ツ)_/¯ Alrighty. 

Obviously this is is already deprecated in Ember 1.13 and removed in 2.0. So if there is a better way to document this let me know.

workmanw:
This has been open for over 3 months at this point. It's not doing anyone any good just sitting in a PR. Is there anything that can be done to move this along? IMHO this is straightforward.

rwjblue:
@workmanw - I'm sorry for the delay here, I have looked at this a few times but have not actually used `Ember.SortableMixin` myself so I'm not terribly well versed in its API and nuances.

I was hoping that @ebryn might be able to take a look...

workmanw:
@rwjblue Okay cool. Thanks for the update. I'd be happy to make any changes that @ebryn et al may suggest.

krisselden:
Looks correct to me, the history of it was the ArrayProxy hooks didn't actually support composability so we moved toward computed properties, but what you would mixin this into is also deprecated.

locks:
Thanks @krisselden, I'm merging it in that case :)

PR #2411, opened 2015-11-16
Fixed a misnaming of the getowner polyfill

acorncom:
/cc @rwjblue


rwjblue:
WHoops!  Good catch, thank you!

PR #2415, opened 2015-11-17
Add interactive intelligence to list of ember users

toddjordan:


kaddae:
email received CC @wifelette

PR #2420, opened 2015-11-20
Fix 2.3-beta.1 changelog link in 2.2 release blog post

knownasilya:
Was pointing to the Ember.js changelog, not the ED changelog as expected.


bmac:
Thanks @knownasilya

PR #2421, opened 2015-11-24
Add 1.13.7 Deprecations to deprecation guide

toddjordan:
working through issue https://github.com/emberjs/website/issues/2256

These are the remaining 1.13.7 deprecations

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -1296,6 +1296,40 @@ component.
 We replaced `tagName="ul"` part by just surrounding the `{{each}}` block with
 `<ul>` tags.
 
+#### Using the `{{with}}` Helper with the `controller` option
+
+Another option deprecated due to the de-emphasis of controllers is the controller
+option used with the `{{with}}` helper.  In prior versions it was possible to
+specify a controller when using `{{with}}`, which would use an instance of the
+specified controller as the context within the block.
+
+```javascript
+{{#with item.posts controller="myposts"}}
+  {{!- numPosts fetched from the controller instance "myposts" }}
+  There are {{numPosts}} posts.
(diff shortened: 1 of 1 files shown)


source/deprecations/v1.x.html.md
```diff
@@ -1339,3 +1373,48 @@ export default Ember.Helper.helper(function([firstArg, secondArg], hash) {
...
+  //fired whenever a blog post is added or removed
+});
+```
+
+#### Overriding `render` When Extending a Component or View.
+
+The `render` method on components and views should not be overridden and will
+go away in Ember 2.x.  Modifications to Ember rendering should be made by
+overriding Ember's new
+[component lifecycle hooks](http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_component-lifecycle-hooks)
```
toddjordan: Pointing users overriding render to the new lifecycle hooks, but would like validation on whether this is an appropriate alternative to point them at.

wycats: I consider the lifecycle hooks, but not the parameters, to be a public API. I thought someone documented them more officially?

toddjordan: Best I could find was here: http://guides.emberjs.com/v2.2.0/components/the-component-lifecycle/ which doesn't include the new stuff.
I remember @rwjblue mentioning there was some uncertainty on the final form of these, which is why I think they haven't been more formally documented.  Maybe he can comment

wycats: The current story is that the hooks, but not their parameters, are fully public. Feel free to PR that!

toddjordan: awesome.  I'll look into it.  Thx!

PR #2425, opened 2015-12-03
Don't suggest to install ember globally.

wolframkriesing:
Suggest not to install ember globally but just for you project locally, to prevent version problems with multiple projects and allowing upgrading individually.

### DIFF
--- a/source/about.html.erb
+++ b/source/about.html.erb
@@ -33,7 +33,7 @@ title: "A framework for creating ambitious web applications."
 <div id="download">
   <div id="download-ember">
     <% highlight("shell", "install-ember-cli") do %>
-npm install -g ember-cli
+npm install ember-cli --save-dev
 ember new my-app
     <% end %>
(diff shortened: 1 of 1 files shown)


rwjblue:
When installed globally ember-cli properly defers to whichever version of ember-cli the current project is using, and only falls back to using the global version if you are not inside of an existing ember-cli project. 

There are no issues with having a different version in multiple projects, or having different global vs local versions.

wolframkriesing:
Everywhere it's suggested to prevent installing any npm package globally, from what you answer this is different in ember?

rwjblue:
There are many packages that are supposed to be installed globally.  `grunt-cli` and `gulp` are a couple examples of CLI based packages that recommend usage globally. They both also happen to have similar functionality to what I described above: they use the locally installed version of the package in a project.

wolframkriesing:
thanks for clarification

PR #2426, opened 2015-12-03
Add 1.13.2 Deprecations to Deprecations Guide

toddjordan:
Part of Issue https://github.com/emberjs/website/issues/2256

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -1023,6 +1023,36 @@ proxying behavior is being used.
 
 Added in [PR #11476](https://github.com/emberjs/ember.js/pull/11476).
 
+#### Using `@guid` and `@item` as key in `{{each}}`
+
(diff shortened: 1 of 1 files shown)


source/deprecations/v1.x.html.md
```diff
@@ -1023,6 +1023,42 @@ proxying behavior is being used.
...
+```
+and
+```hbs
+{{#each people key='@item' as |person|}}
+...
+{{/each}}
+```
+with
+```hbs
+{{#each people key='@identity' as |person|}}
```
rwjblue: I would prefer to just include the latter example (the `simply` one).

rwjblue:
left one comment, but looks great otherwise!

source/deprecations/v1.x.html.md
toddjordan: agreed, will update

toddjordan:
fixt 👍

PR #2428, opened 2015-12-06
Document Remaining Deprecations for 1.13.5

toddjordan:
Part of the fix for https://github.com/emberjs/website/issues/2256


source/deprecations/v1.x.html.md
```diff
@@ -1172,6 +1172,62 @@ Object.freeze(c);
...
+  }
+});
+
+const copy = Obj.create().frozenCopy();
+```
+
+with:
+
+```javascript
+const copy = Object.freeze();
```
mixonic: I expect you want:

```js
'use strict';

let a = {};
Object.isFrozen(a); // => false
Object.freeze(a);
Object.isFrozen(a); // => true
a.foo = 'baz'; // => throws TypeError
```

It may also be worth pointing to the documentation for [Object.seal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal) and [Object.freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze)

toddjordan: Thanks for the feedback!  Will update.

source/deprecations/v1.x.html.md
```diff
@@ -1172,6 +1172,68 @@ Object.freeze(c);
...
+a.set('foo', 'baz'); // => throws TypeError
+```
+
+#### Using `/` for namespace in the `{{render}}` helper
+
+Identifying controllers not under the application namespace no longer should take
+a slash as the separator, as Ember has settled on using dot notation for namespaces.
+
+Instead of using `{{render 'blog/posts'}}`, use `{{render 'blog.posts'}}`.
+
```
wifelette: Can you take another stab at this? @wycats was just saying he didn't quite feel it was clear enough.

source/deprecations/v1.x.html.md
```diff
@@ -1172,6 +1172,68 @@ Object.freeze(c);
...
+#### Calling `Ember.set` and `Ember.get`
+
```
wifelette: How about "Non-Standard Use of Calling..."—we don't want to accidentally indicate that we've completely deprecated `Ember.set` and `Ember.get`.

wifelette:
Should be good to merge once the comments are addressed :)

source/deprecations/v1.x.html.md
toddjordan: k working on it

toddjordan:
Updated PR with changes to address comments.

wifelette:
Almost there! 

Gotta resolve the current conflicts so it'll merge cleanly, but also, @wycats was concerned there might still be perfecting work to be done on the "Using `/` for namespace in the `{{render}}` helper" section. I pinged @rwjblue to hopefully take a look tomorrow, and see if he agrees and has ideas, or if he thinks this is good to merge.

Gaurav0:
@wifelette @wycats @rwjblue Ping... please review and merge!

PR #2432, opened 2015-12-18
Fix blog post reference to Ember instead of ED

### LINKED ISSUE #2417: Ember Data 2.2 blog post confusing
I would suggest that you refrain from talking about the Ember and Ember Data in the same blog post causes too much confusion.

The blog post here is really confusing. Maybe it is just the changelog reference that is wrong?

http://emberjs.com/blog/2015/11/17/ember-data-2-2-released.html

ember_js_-_ember_data_2_2_and_2_3_beta_released

michaelrkn:
Fixes #2417.


acorncom:
@wifelette Just noticed this.  Happy to help merge pull requests if you want, or can stay on the issues side of things if you'd prefer ...

PR #2433, opened 2015-12-21
document remaining deprecations for 1.13.0 and 1.13.1

toddjordan:
Should complete fixes for https://github.com/emberjs/website/issues/2256

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -714,6 +714,13 @@ nearly always be replaced by proper use of data passing and block params. See
 the guide on [differences in yielded blocks](http://emberjs.com/deprecations/v1.x#toc_differences-in-yielded-blocks)
 for a complete example of using block params over the `{{view}}` keyword.
 
+
+##### `view` and `viewClass` arguments to `{{outlet}}`
+
+Since the View API is deprecated starting in 1.13, providing the `view` or `viewClass` argument to the `{{outlet}}`
+helper is likewise deprecated in favor of migrating to component-based approaches, as explained earlier in this
+section.
+
 #### Ember.LinkView
 
 As a consequence of the deprecation of `Ember.View`, many internal views have
@@ -939,7 +946,7 @@ Thing.reopenClass({
 
 #### Ember.ReduceComputed / Ember.ArrayComputed
 
-This two little monsters served us well for a long time. They provided semantics that allowed to
+These two little monsters served us well for a long time. They provided semantics that allowed to
 perform in-place modifications of an existing array instead of replacing the entire array like the
 regular `map/reduce` methods in javascript do, which was necessary to avoid expensive repaintings when
 rendering lists with the `{{each}}` helper.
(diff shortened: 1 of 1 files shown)


source/deprecations/v1.x.html.md
```diff
@@ -1260,6 +1324,18 @@ Ember 1.13 deprecated `bind-attr` in favor of the new syntax.
 
 To aid in the migration, you can use the [legacy-bind-attr plugin](https://github.com/emberjs/legacy-bind-attr) to restore this behavior in Ember 2.0 and silence the deprecation warning.
 
+#### Using `this.get('template')`
+
+Prior to 1.13, you developers had to check for the existance of the internal `template` property to determine the existance of a yielded block.
```
acorncom: Maybe drop the `you` before `you developers`?

toddjordan: sorry, meant to be "your developers", but agreed that just "developers" would be best

source/deprecations/v1.x.html.md
```diff
@@ -951,6 +958,52 @@ over the native counterparts.
 They will be extracted as an ember addon in case you need to keep using them, but the recomendation now is to
 just use the native array methods or those in libraries like Underscore/Lodash.
 
+#### Ember.ReduceComputedProperty / Ember.ArrayComputedProperty
+
+In addition to `Ember.ReduceComputed` and `Ember.ArrayComputed` you could add a property function call, making use of
+the ReduceComputedProperty and ArrayComputedProperty classes.
```
wifelette: How about "you _were_ able to add a property..."

source/deprecations/v1.x.html.md
```diff
@@ -951,6 +958,52 @@ over the native counterparts.
...
+    return totalValue + newValue;
+  },
+  removedItem(totalValue, newValue) {
+    return totalValue - newValue;
+  }
+}).property('values')
+```
+
+Now that these APIs are deprecated, use the native JavaScript reduce along with the regular computed API:
+
```
wifelette: We should also point people to the Ember.Computed family of functions—http://emberjs.com/api/classes/Ember.computed.html.

@stefanpenner Should we have added `Ember.computed.reduce`?

source/deprecations/v1.x.html.md
```diff
@@ -951,6 +958,52 @@ over the native counterparts.
...
+```js
+  didInsertElement() {
+    this._super(...arguments);
+    ...
+    Ember.run.schedule('afterRender', () => {
+      this.set('myValue', value);
+    });
+  }
+```
+
```
wifelette: The correct new approach is to use `didRender`

If that doesn't work for some reason, we should fix it.

In what situations do you need to use `afterRender` in a `didInsertElement` hook in the first place?

source/deprecations/v1.x.html.md
```diff
@@ -1172,6 +1230,12 @@ Object.freeze(c);
 c.swapNames();  // EXCEPTION
 ```
 
+#### Ember.EnumerableUtils
+
+`EnumerableUtils` has been deprecated in favor of native implementations. You can consult the
+[JavaScript Array documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)
+to find suitable replacements for `EnumerableUtils` functionality.
+
```
wifelette: We might want to suggest `underscore` or `lodash` if the functionality a user wants isn't available in the native implementations.

wifelette:
Made a bunch of comments, but this should otherwise be good to go once they're addressed. 

This is AWESOME. Cannot _wait_ to close the damned original issue :stuck_out_tongue_closed_eyes:

source/deprecations/v1.x.html.md
toddjordan: There's a deprecation for setting properties in `didRender` as well.  I'll find some of the links/issues that have explained.

I can't say that there are great reasons for setting properties in didInsertElement, but I do know that it is done with existing codebases.  If an app has code that is in production that is setting values within a didRender, this is the way to work around it afaik.

wycats: We're pretty opposed to people using the `afterRender` queue, because Ember can't control the exact timing of `afterRender` callbacks relative to each other, while we can with `didRender`.

toddjordan: yeah, get that.  I will remove the afterRender.   I've seen it as a useful tool in a bind, but agree we shouldn't lead folks there off the bat.

toddjordan:
Updated PR with changes to address comments.

PR #2457, opened 2016-01-27
Remove the San Juan, PR meetup

gcollazo:
This meetup slowed down and finally became
inactive. I wrote a blog post about the reasons
but in summary: our island is too small for a 
framework/language specific meetup.

We started a new meetup aimed at the general
developer community an it seems to be doing a lot
better. There's a lot of interest about Ember.js
so we'll keep doing talks and hand on learning
events.


wifelette:
Hi @gcollazo —

That make sense, indeed, something more general would likely work better with a smaller developer group.

One thing to consider is interspersing remote talks with your regular content. I know it's less engaging, but it can help with locations that might be short on speakers / presenters.

gcollazo:
@wifelette We are thinking about doing remote talks. Perhaps I'll bother you later with speaker suggestions or contacts. Thanks a lot for the support and see you at EmberConf :smile:

PR #2459, opened 2016-02-02
Add Startaê logo to the Who's using Ember page

renatocarvalho:
screen shot 2016-02-02 at 12 12 21 am


wifelette:
Hiya—

Can you confirm, am I correct that this is a consulting firm that uses Ember for client projects?

renatocarvalho:
@wifelette Yes! We're a consulting firm that uses Ember for both client and internal projects.

PR #2461, opened 2016-02-03
Emphasize style attribute binding should not be quoted

balinterdi:
I've spent quite some time debugging this (see [my blog post](http://balinterdi.com/2016/02/03/binding-style-attributes-warning-in-ember.html)), and although the section in the guides is correct, I think others would benefit by having this emphasized.

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -345,9 +345,12 @@ making the content a `SafeString`. For example:
 ```js
   myStyle: Ember.computed('color', function() {
     var color = escapeCSS(this.get('color'));
-    return new Ember.Handlebars.SafeString("color: " + color);
+    return Ember.String.htmlSafe("color: " + color);
   })
 ```
+
+Make sure you don't put quotes around the sanitized string, `myStyle`, when you
+bound it in the template. This would prevent Ember from seeing it as safe.
 
 You can learn more about `SafeString`s and writing code that prevents XSS
 attacks by reading the [Writing
@@ -1001,7 +1004,7 @@ In many cases you can move your logic later in the component lifecycle by implem
     this._super(...arguments);
     this.set('myValue', value);
   }
-``` 
+```
 
 #### beforeObserver
 


chancancode:
Seems like a bug

cc @wycats

wycats:
@balinterdi This definitely seems like a bug. `concat` should do the escaping, and produce a safe string.

For what it's worth, the exact semantics of property vs. attribute are somewhat undefined at the moment, so I'm pretty nervous about having people depend on a side effect of property vs. attribute that provides even-less-defined safe string semantics.

balinterdi:
Should I create a proper issue for this in the emberjs repo? I can include the example in the blog post (which I have also extracted to a gist).

source/deprecations/v1.x.html.md
```diff
@@ -349,6 +349,13 @@ making the content a `SafeString`. For example:
   })
 ```
 
+Make sure you don't put quotes around the string marked-as-safe when you bound
+it in the template, as this would prevent Ember from seeing it as safe:
+
+```handlebars
+<div style={{myStyle}}></div>
```
mixonic: This is exactly the same example as a few paragraphs above. Perhaps you can just add test to one of those paragraphs making this more explicit? Repeating the code doesn't seem helpful.

mixonic:
This is definitely needed right now. We distrust anything that goes through `concat`, so anything quoted.

source/deprecations/v1.x.html.md
balinterdi: True, I've moved that paragraph under the definition of the `myStyle` property and removed the duplicated template code snippet.

balinterdi:
@mixonic I'm not sure I see what you mean, what is needed? Do you consider the quoted version triggering the warning a bug?

mixonic:
@balinterdi sorry for the confusing note- I'm not sure I consider it a "bug". Two safe things, when combined, may be unsafe. However I'm sure we can do better in common cases like `style="{{foo}}"`.

Regardless I'd like to merge this docs improvement. @balinterdi can you also change the JS part of the example to:

```js
return Ember.String.htmlSafe("color: " + color);
```

instead of using the Handlebars namespace? If you do this I'll :shipit:!

balinterdi:
Thank you, just changed it to use `Ember.String.htmlSafe`, which I actually also prefer.

mixonic:
Thanks @balinterdi

PR #2462, opened 2016-02-07
add Phoenix, AZ meetup

sstgithub:


wifelette:
Heya!

This is great. Can you update the photo link to point to a more cropped square headshot? IIRC, the existing code won't do anything to the image, and this large one could end up looking like a big grey box with a small dot in middle :p Your GitHub gravatar seems to be cropped, that could work well too.

PS: You're not being singled out! It looks like there are a number of photos that ideally would be replaced with proper headshots, so it's not just you. I'll start working on replacing them, just happen to be starting now and your PR came in :)

sstgithub:
Hmm I was kinda going for that whole abstract small-dot-in-the-middle-of-a-big-grey-box look but okay, I’ll go ahead and change it :p

sstgithub:
Changed to 150x150 headshot. Let me know if this works, and thanks for pointing this out!

sstgithub:
Also wouldn't adding a `style="width: 150px"` to line 53 of `meetups.js` fix this problem for everyone?

wifelette:
I thiiiink that would just scale the photo, not crop it appropriately :p Plus it would likely lead to folks linking to giant photos that then take forever to load. I think we're on manual crop-work for now :stuck_out_tongue_winking_eye:

PR #2463, opened 2016-02-07
Use GitHub Pagination for the loading example

zebMcCorkle:
Currently, the example grabs all of Ember's PRs over the network then splices out all but the 3 most current. GitHub's API has pagination, so this isn't necessary and just increases network load.


rwjblue:
Awesome, thank you!~

PR #2468, opened 2016-02-18
Fixing broken instance initializer deprecation link

### LINKED ISSUE #2306: Instance initializer deprecation guide link is broken
There is a link in http://emberjs.com/deprecations/v1.x/#toc_access-to-instances-in-initializers
to http://emberjs.com/deprecations/v1.x/deprecations/1.11/instance-initializers which results in 404

green-arrow:
I believe Robert Jackson attempted to fix this a while back by changing the link path and the file name, and this works fine locally, but the link is still broken on the deployed website. I believe this is due to the file extension not being `.html.md`.

Fixes #2306


rwjblue:
Aha!  OK, thank you!

PR #2469, opened 2016-02-22
Add Ember-CLI welcome script

acorncom:
This file is going to be referenced from the Ember-CLI `application.hbs` file.  We're trying to achieve a few things here:

- provide a nice 'getting started' experience for folks new to the Ember ecosystem
- keep from unnecessarily cluttering the Ember default template in any way that would make upgrades more painful for folks already using Ember

We ( @mixonic and I ) are hoping that referencing this script from the Ember-CLI `application.hbs` will be a reasonable compromise.

Before:
before

After:
screen shot 2016-02-21 at 8 40 30 pm

Questions:
- Is this the text that we want to put in? Is it warm and friendly enough?  Representative of the Tomster way of saying things? :smile:  Anything we are leaving out?  
- Do we want to link to anything else?  I've heard recommendations to link to the Ember Inspector, but was trying to avoid linking to anything specifically called out in the Guides and also avoid linking to Slack so we don't run into size problems there ...  Plus, linking to specific sub-sections of the guides requires additional 'version' info as we don't currently have a `/current/` url yet ;-)
- I'm assuming we'd want to localize this for different languages.  That seems like a v2 of this, but any thoughts on how to handle that?

Additional thoughts:
- we decided to skip worrying about the 'offline experience' of this as the only time folks would see this page would be immediately after an `npm install` (unless this is shown while at a conference without Internet access, in which case the page should load as normal but throw a 404 behind the scenes due to the inability to load the external script)

/cc @locks @wifelette @mixonic @tomdale

### DIFF
--- a/source/javascripts/ember-cli-welcome.js
+++ b/source/javascripts/ember-cli-welcome.js
@@ -1,0 +1,88 @@
+// Used as part of the initial Ember-CLI application.hbs template to welcome new folks
+// This is fairly visible, so change with care! :-)
+
+jQuery('#title').hide();
+
+jQuery('head').append('<style>' +
+  'body {' +
+  '  font-family: "Helvetica Neue", "Helvetica", "Roboto", "Arial", sans-serif;' +
+  '  font-size: 16px;' +
+  '  line-height: 1.35em;' +
+  '  background: #FFFBF5;' +
+  '}' +
+  'p {' +
+  '  margin: 0 0 .75em;' +
+  '}' +
+  'a {' +
+  '  color: #dd6a58;' +
+  '  text-decoration: none;' +
+  '}' +
+  'a:hover {' +
+  '  color: #c13c27;' +
+  '}' +
+  'h2 {' +
+  '  color: #dd6a58;' +
+  '  margin-top: 1em;' +
+  '  font-size: 1.75em;' +
+  '}' +
+  '.welcome-box {' +
+  '  width: 800px;' +
+  '  color: #865931;' +
+  '  margin: 2em auto 0;' +
+  '}' +
+  '.welcome-box img {' +
+  '  float: left;' +
+  '  margin-right: 3em;' +
+  '  width: 300px;' +
+  '  height: 336px;' +
+  '}' +
+  '.welcome {' +
+  '  width: 450px;' +
+  '  float: left;' +
+  '}' +
+  '.welcome ul {' +
+  '  list-style: disc;' +
+  '  padding-left: 2em;' +
+  '  margin-bottom: .75em;' +
+  '}' +
+  '.welcome > ul > li {' +
+  '  padding-bottom: .5em;' +
+  '}' +
+  '.postscript {' +
+  '  clear: both;' +
+  '  text-align: center;' +
+  '  padding-top: 3em;' +
+  '  font-size: 12px;' +
+  '  color: #aaa;' +
+  '  font-style: italic;' +
+  '}' +
+  '.postscript code {' +
+  '  background-color: #F8E7CF;' +
+  '  border-radius: 3px;' +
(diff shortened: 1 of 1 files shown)


tomdale:
This is awesome! I've always wanted to have a nice experience for the default app when running, so I'm glad you all are tackling this.

One thing that might be nice is to strike a more congratulatory tone. Your first Ember app is an auspicious moment—we should make sure users feel like we're acknowledging how exciting it is.

The links to reference material is great, but we may also want to consider linking back to the GSG. Something like "Is this your first Ember app? We have a tutorial that walks you through everything you need to know."

I'm fine with punting on the offline experience for now. One way of dealing with it would be to write it to localStorage and prefer that if available—but it's definitely an edge case.

Turbo87:
as mentioned in https://github.com/ember-cli/ember-cli/pull/5501 it would be better to extract this page into an `ember-getting-started` addon. unless I'm missing something this would require five steps:

- [ ] create `ember-getting-started` addon
- [ ] add `app/templates/application.hbs` with the proposed contents
- [ ] add necessary CSS to the addon
- [ ] add `ember-getting-started` to the default `app` blueprint in `ember-cli`
- [ ] remove `application.hbs` from the default `app` blueprint in `ember-cli`

locks:
@Turbo87 https://github.com/locks/ember-welcome-page

Should I publish it under my own name?

acorncom:
Regarding congratulatory tone, I've posted a few other text options on UseCanvas.  @tomdale any feedback on these? https://usecanvas.com/ember-docs/welcome-drafts/33KmyOjhfBdrTmJeIx1dn4  We can mix and match between the options or iterate on one of them if it's in the right ballpark but still not coming across the way you intend ...

mixonic:
@rwjblue and I have discussed this in person. He feels confident we can spike an addon that would detect if an app exists, and if it does not, then display this prompt.

"detect if an app exists" consists of a set of checks:

* Has an `application.hbs` been created?
* Has an `index.hbs` been created?
* Have any routes been created?

These are all things not too hard to check for. Then we would not even need to pull in a script.

locks:
@mixonic you mean instead of what I'm doing in the repo linked above? send over a PR, please :)

Turbo87:
I just tested @locks `ember-welcome-page` addon and it works quite well.

There is one minor issue if you try to overwrite the `application.hbs` without removing the addon first it will complain that the `application.js` file already exists. I'm not sure if that is actually a problem though because the welcome page should explain how to remove the welcome page addon.

Another thing that might be problematic for power users is that they now have to manually remove the addon each time they create a new app, while previously they just had to overwrite or delete the existing `application.hbs` or just start working and adjust the file later. This could possibly be solved by something like a `--no-welcome-page` flag on the `ember new` command though.

acorncom:
@locks re having this repo under your name, I think it makes more sense for it to be an Ember repo so that others can be assigned write access via groups in the future ...

locks:
@acorncom I meant if I should publish temporarily under my account to get it moving on the CLI side. I agree and I'll transfer the repo as soon as approved :)

acorncom:
:+1:

Turbo87:
@locks are we talking NPM publish or GitHub publish?

locks:
npm. I guess it's not necessary, we can use github dep, right?

Turbo87:
@locks yes, as long as we're still experimenting with it a github dep should be sufficient. once the addon is somewhat "finished" the PR to `ember-cli` should use a NPM dep though.

acorncom:
I'm closing this PR as we've moved development to the `ember-welcome-page` repo.  Feedback or comments on our v1 issue (https://github.com/locks/ember-welcome-page/issues/4) are more than welcome!

Issue #2470, opened 2016-02-22
Silicon Valley Meetup has broken link


ef4:
The link from emberjs.com to the Silicon Valley meetup is dead. 

cc @ebryn

wifelette:
Unsure if it's just the site that's not maintained, or the _group_. If it's just the site, we can link to here: http://www.meetup.com/Silicon-Valley-EmberJS-Meetup/

PR #2471, opened 2016-02-24
Give Canadian meetups province-level information

backspace:
If American cities show their states, Canadian cities might as well show their provinces.

I also threw in a fix for [the broken Silicon Valley link](https://github.com/emberjs/website/issues/2470).


wifelette:
Good stuff, thanks!

PR #2473, opened 2016-02-27
Add missing lat long for meetup in Riverside, CA

ryanlabouve:
The missing lat / long data was causing the link on http://emberjs.com/community/meetups/ to go directly to the meetup's page instead of interacting with the google map (like the other links do). 

I pulled the lat/long from the address they typically host their startup from as seen on: http://www.meetup.com/Ember-Riverside-Meetup/events/224108684/, or directly on the image http://photos1.meetupstatic.com/photos/event/9/d/8/2/600_371440322.jpeg

I also fixed the url for the organizer's profile image.


locks:
Appreciated!

PR #2477, opened 2016-02-29
Port Responsive Header and Footer from Guides

ryanlabouve:
I'm working on the 2016 Ember Community Survey landing page. It would be nice to have a responsive header since likely many of the folks viewing the survey on their mobile devices.

I did a quick port of the header from https://guides.emberjs.com/v2.3.0/ across the whole website.

Would something like this be welcomed / helpful?

A few notes:
* The guides website is using bourbon / neat and the current ember site is not.
* My port involves taking some of the compiled CSS from the guides to avoid affecting many base styles and creating potentially unwanted visual side effects on the rest of the site. 
* If a more aggressive approach would be preferred (i.e. porting all of the base styles from guides and switching) I'd be happy to assist in any way helpful.

Screenshot:
ember_js_-_a_framework_for_creating_ambitious_web_applications_

PR #2484, opened 2016-03-02
Community Survey 2016 Landing Page

ryanlabouve:
Here's the first shot at the landing page for the 2016 community survey!

Would be nice to get the responsive header with this so that the layout would break down properly on mobile: https://github.com/emberjs/website/pull/2477

screenshot


wifelette:
As per docs discussion, @ryanlabouve and @mixonic are going to take a final pass on this to match the branding more to the Ember stuff than the EmberConf stuff :)

mixonic:
Now more like

PR #2486, opened 2016-03-04
Fixes https://github.com/emberjs/guides/issues/1302

locks:
Fixes https://github.com/emberjs/guides/issues/1302.

Guides version was hardcoded into the path.
I have added logic to fetch the latest version from the guides'
versions.json.

### DIFF
--- a/source/layouts/guide.erb
+++ b/source/layouts/guide.erb
@@ -1,10 +1,14 @@
 <!DOCTYPE html>
 <html lang="en">
   <%
+    require 'json'
+    require 'open-uri'
+
+    @version = JSON.parse(open('https://guides.emberjs.com/versions.json').read).last
     @path = current_page.url.gsub(/^\/guides/, '')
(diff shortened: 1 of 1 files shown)


source/layouts/guide.erb
```diff
@@ -1,10 +1,14 @@
 <!DOCTYPE html>
 <html lang="en">
   <%
+    require 'json'
+    require 'open-uri'
+
+    @version = JSON.parse(open('https://guides.emberjs.com/versions.json').read).last
```
acorncom: @locks Instead of reading this from a web page, couldn't we just read it from the local file on build?

locks: what local file?

acorncom: @locks Sorry, thought this was the guides repo, carry on ;-)

locks: :+1:

stefanpenner:
Good catch

PR #2496, opened 2016-03-11
Fix spacing issues on footer to match guides

ryanlabouve:
This fixes the spacing issues in the footer. 

The only issue I'm not address is that that guides is using the typeface _Proxima Nova_ while the website is using _Helvetica Neue_. This causes very slight horizontal shifting (as you can see in the GIFs).

If we want to switch the default typeface to _Proxima Nova_ to match guides, that would be spiffy! And I'd be happy to help! However, I figured this tweaks would be useful either way.

Here are the comparison GIFs:
---------------
Here's for the left-matter:
_before:_
before-type-in-info3

_after (i.e. in this PR):_
type-in-info

----------------
In the social icons:
_before:_
before-type-in-icons

_after:_
type-in-icons
--------------
Finally, here's what mobile looks like:

mobile-side-by-side

/cc @wifelette


locks:
Hm, why does g+ have two different icons?

ryanlabouve:
Great question. Sorry I forgot to mention that.

I know for sure that the website is using font awesome (via fontello). I _think_ guides is as well, but perhaps a newer version.

To further complicate matters, the latest version of font awesome is using an updated logo for g-plus (http://fortawesome.github.io/Font-Awesome/icon/google-plus/):
font_awesome_icons

Options:
1. Update guides and website to use newest version of font awesome
2. Match the website to what guides has currently (maybe looking through older versions of font awesome and making a new fontello pack?)
3. Punt for now

locks:
I'd like to get 1. done if possible! :)

ryanlabouve:
Sure thing! I'll update the icons on both projects, and link the other guides PR back here.

ryanlabouve:
@locks updated google plus icon in both projects!

mixonic:
@homu r+

mixonic:
I am homu, hear me roar

acorncom:
<groan>

PR #2501, opened 2016-03-16
Fix typos on deprecation page

sbatson5:
Found some small typo's on the deprecations page


source/deprecations/v1.x.html.md
```diff
@@ -1212,7 +1212,7 @@ Or make all inputs to the helper unbound:
 
 #### Ember Freezeable
```
locks: forgot to correct here :]

sbatson5: Thanks!  Revised.

mixonic:
Thanks @sbatson5

PR #2503, opened 2016-03-18
Add survey reminder to layout — WIP

ryanlabouve:
This adds a simple reminder bar to the website to remind folks to fill out their 2016 survey. Preview below.

@mixonic can you give me some direction on the copy? Also, should I use a hard date as the stopping point or a little Javascript for something like "2 days left"?

Preview:
ember_js_-_a_framework_for_creating_ambitious_web_applications_

Mobile (responsive page): 
9bby70rq8ciaaaaasuvork5cyii

Mobile (unresponsive page):
ember_js_-_a_framework_for_creating_ambitious_web_applications_ 2

### DIFF
--- a/source/_survey-reminder.html.erb
+++ b/source/_survey-reminder.html.erb
@@ -1,0 +1,9 @@
+<% if !current_page.data.no_survey_reminder %>
+  <div class="survey-reminder-wrapper">
+    <div class="survey-reminder-container">
+      Only a <strong>4 days left</strong> to
+      <% link_to '/ember-community-survey-2016' do %>
+        submit your community survey response<% end %>!
+    </div>
+  </div>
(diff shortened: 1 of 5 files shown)


acorncom:
@ryanlabouve I like the idea of a little JS for "two days left" :+1: If you did that, you might also be able to remove the reminder bar once the date is passed so we don't have to remove it exactly on the day it expires ...

mixonic:
I've pushed a version that uses "4 days left" and some alternative copy on the survey page itself. I look at it in a variety of layouts and things look good. I'm going to move forward with this low-tech version, we can always enhance it over the weekend.

Monday is our last survey day, so I want to get this live even if we want tweaks. :tada:

PR #2506, opened 2016-03-18
Fixes incorrect relative path if viewed at /about instead of /

### LINKED ISSUE #2505: 404 error -- about.html.erb#L67 emberconf-banner.png
http://emberjs.com/about/ emberconf-banner img src points to incorrect relative path:
http://emberjs.com/about/images/about/emberconf-banner.png

https://github.com/emberjs/website/blob/master/source/about.html.erb#L67

acorncom:
Fixes #2505


acorncom:
@mixonic would you mind merging / deploying this fix?

PR #2508, opened 2016-03-18
Fix typo in survey reminder header

ryanlabouve:
@acorncom pointed out a typo on the guides survey reminder that was also on website. Thanks @acorncom!

### DIFF
--- a/source/_survey-reminder.html.erb
+++ b/source/_survey-reminder.html.erb
@@ -1,9 +1,9 @@
 <% if !current_page.data.no_survey_reminder %>
   <div class="survey-reminder-wrapper">
     <div class="survey-reminder-container">
-      Only a <strong>4 days left</strong> to
+      Only <strong>4 days left</strong> to
       <% link_to '/ember-community-survey-2016' do %>
-        submit your community survey response<% end %>!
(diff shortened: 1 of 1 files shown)


source/_survey-reminder.html.erb
```diff
@@ -1,9 +1,9 @@
 <% if !current_page.data.no_survey_reminder %>
   <div class="survey-reminder-wrapper">
     <div class="survey-reminder-container">
-      Only a <strong>4 days left</strong> to
+      Only <strong>4 days left</strong> to
       <% link_to '/ember-community-survey-2016' do %>
-        submit your community survey response<% end %>!
+        submit your community survey&nbsp;response<% end %>&nbsp;!
```
mixonic: Why these nbsps?

Certainly the one before the `!` can't be right?

ryanlabouve:
@mixonic the `&nbsp`'s prevent typographic orphans (i.e. you don't have a `!` floating alone on another line on narrow screens).

The `&nbsp;` between the link tag and the `!` is for visual spacing. With the underline on the link, it looks strange to have the `!` run into the underline. Let me grab some screenshots.

mixonic:
dropped the last one and I'll roll it out.

PR #2510, opened 2016-03-18
Adding B Lab to Ember Users list

jonpitch:


wifelette:
Heya—

Could you update the blurb to be a bit more specific? 

> We use Ember to build our app/product which helps people use business as a force for good.

...would work just as well, if it still sounds correct :)

jonpitch:
@wifelette no problem, I'll do that momentarily.

wifelette:
Thanks!

PR #2511, opened 2016-03-18
Add conde nast logo

dlopezh:


wifelette:
Looks like this needs to be rebased :)

dlopezh:
@wifelette done :)

PR #2515, opened 2016-03-21
Fix EmbeddedRecordsMixin import path

kevinansfield:
The embedded records mixin is located at `'ember-data/serializers/embedded-records-mixin'` not `'ember-data/mixins/embedded-records'` as I found out when copy/pasting the example. This fixes the example code and adds a reference to the pull request that introduced the changes.


bmac:
Thanks @kevinansfield.

PR #2517, opened 2016-03-22
Remove survey reminder

ryanlabouve:
This years 2016 Ember Community Survey is now closed!

Removing the survey reminder bar that was added temporarily below the header.

On the actual survey page :
* Changed the text for it to notify that the survey is closed
* Removed button that links to survey

@mixonic should I also remove the partial itself for the survey reminder?


locks:
As I commented on the related Guides PR, I think this would be a useful thing to put in the style guide, and remove it from the repositories for now.

mixonic:
@ryanlabouve we should remove everything IMO including the CSS

PR #2527, opened 2016-03-29
Fix survey graph label

kpfefferle:
The vertical axis on the first graph "How long have you been working with Ember?" is labeled "Months", but it seems that it should be "Percent".


kpfefferle:
cc: @mixonic

mixonic:
Thanks @kpfefferle !

PR #2532, opened 2016-04-03
Use github raw instead of guides.emberjs.com because my system sucks.

rwjblue:
I have updated the root SSL certs (I'm still on Mavericks) via both system update and RVM's `rvm osx-ssl-certs update all`, but neither helped.

This works, but feels odd....


rwjblue:
Without this (or otherwise hand editing the guides layout file), I can't deploy the site :crying_cat_face:

PR #2535, opened 2016-04-07
Documentation for deprecation of model param in `{{render` helper

rob-e-pitts:
Documents [this PR](https://github.com/emberjs/ember.js/pull/13268) originating from [this issue](https://github.com/emberjs/ember.js/issues/13183). I presume this change will be in 2.5.x so I documented it as such. I can obviously change this though.

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -375,3 +375,22 @@ export default Ember.Component.extend({
   }
 });
 ```
+
+### Deprecations added in 2.5
+
+#### Model param in `{{render` helper
+
+Using the model param in the `{{render` helper is deprecated in favor of using
+components. Please refactor to a component and invoke thusly:
+
+Before:
+
+```handlebars
+{{render "foo-bar" someModel}}
+```
+
+After:
+
+```handlebars
+{{foo-bar model=someModel}}
+```


acorncom:
@rwjblue per your comment here: https://github.com/emberjs/ember.js/pull/13268#issuecomment-206504511 the deprecation should be for 2.6, correct? If so, this says 2.5 :smile:

rob-e-pitts:
Ack! Thank you.

rwjblue:
Confirm, 2.6 (I should have :coffee: before merging ;P )

rob-e-pitts:
Thank you! Follow-up PR incoming.

rwjblue:
@rob-e-pitts - I submitted https://github.com/emberjs/website/pull/2536 (had a few others that were wrong too).

rwjblue:
@rob-e-pitts - I was chatting with @mmun about this, and I think we should add a little bit more info on the "how".  

For example, if you had:

```hbs
{{! app/templates/foo-bar.hbs
<p>{{someProp}} template stuff here</p>
```

```js
// app/controllers/foo-bar.js
export default Controller.extend({
  someProp: Ember.computed(function() {
    return 'yolo!';
  })
});
```

Would be refactored to:

```hbs
{{! app/templates/components/foo-bar.hbs
<p>{{someProp}} template stuff here</p>
```

```js
// app/components/foo-bar.js
export default Component.extend({
  someProp: Ember.computed(function() {
    return 'yolo!';
  })
});
```

mmun:
:+1:

rwjblue:
submitted as https://github.com/emberjs/website/pull/2537

rob-e-pitts:
Awesome, thank you.

Issue #2538, opened 2016-04-10
Deprecation Guide for 2.4-LTS items


mixonic:
The deprecation guide is missing two items for Monday's release:

* [x] A deprecation guide for the private legacy view and controller flags.
* [x] A deprecation guide for deprecating `{{#render` https://github.com/emberjs/website/pull/2540

Additionally, we currently lack a unified way to note if a public or private API is being deprecated. Better than flagging things as public/private would be to add the `until` value of a deprecation to guide entries.

mixonic:
![image](https://cloud.githubusercontent.com/assets/8752/14414933/adc5e396-ff67-11e5-8e2f-ae4590a023c0.png)
 via @locks

rwjblue:
Added https://github.com/emberjs/website/issues/2543 for the last point (and I fleshed it out a bit more).

PR #2542, opened 2016-04-11
Add section on deprecating legacy support addons

toddjordan:
Addresses the legacy addon part of #2538


rwjblue:
You rock! Thank you!

PR #2554, opened 2016-04-15
Latest ember.js available on release channel is 2.5.0

Gaurav0:


locks:
Superceeded by #2555 since it also fixes the release date, and Ember Data's information.

rwjblue:
Thanks for spotting this, and working on the fix!

PR #2559, opened 2016-04-17
Update ember-users links to open in a new tab (as its an external webpage)

raikoroje:


rwjblue:
Can you split the adding of feedza from the template change?

PR #2561, opened 2016-04-17
[On Hold] Removed `Ember.merge` from the list of 2.x deprecations. 

workmanw:
Per emberjs/ember.js#13344 it is being undeprecated.

EDIT: Do not merge just yet. We should wait at least until Ember 2.5.1 has shipped with the deprecation removed.

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -363,28 +363,6 @@ Support the following forms will be removed after 2.4:
 {{/render}}
 ```
 
-### Deprecations Added in 2.5
-
-#### Ember.merge
-
-##### until: 3.0.0
-##### id: ember-metal.merge
-
-`Ember.merge` is deprecated in favor of `Ember.assign`, which is now a public API since Ember 2.5. For typical usages, this
-should be a 1-to-1 replacement:
-
-Before:
-
-```js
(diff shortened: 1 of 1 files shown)


workmanw:
I looked at both blog posts, ["EMBER.JS 2.4 AND 2.5 BETA RELEASED"](http://emberjs.com/blog/2016/02/29/ember-2-4-released.html) and ["EMBER.JS 2.4-LTS, 2.5, AND 2.6 BETA RELEASED"](http://emberjs.com/blog/2016/04/11/ember-2-5-released.html), both introduce `Ember.assign`, but neither mention `Ember.merge` as deprecated.

rwjblue:
@workmanw - Since this URL is included in 2.5.0, I think we should wait at least until 2.5.1 is released. I'd hate to have a URL listed that doesn't work.

workmanw:
@rwjblue Ah yes, that's an excellent point. I'll update the subject line to have `[on hold]` in the header. Or would it be better to just nuke this for now and revisit later?

rwjblue:
No, no.  I'm just saying we can't merge until 2.5.1 is released, which should be Monday/Tuesday of this week.

Gaurav0:
2.5.1 has been released.

rwjblue:
Thanks @Gaurav0 !

PR #2562, opened 2016-04-19
Improves the call to action in the frontpage

locks:
Screenshot:

Ember.js home page with new call to action widget

### DIFF
--- a/source/about.html.erb
+++ b/source/about.html.erb
@@ -30,23 +30,19 @@ title: "A framework for creating ambitious web applications."
   A framework for creating ambitious web applications.
 </h1>
 
-<div id="download">
-  <div id="download-ember">
-    <% highlight("shell", "install-ember-cli") do %>
-npm install -g ember-cli
-ember new my-app
-    <% end %>
-  </div>
+<div class="call-to-action">
+  <div class="top">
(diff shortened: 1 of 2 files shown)


source/about.html.erb
```diff
@@ -30,23 +30,19 @@ title: "A framework for creating ambitious web applications."
...
+<div class="call-to-action">
+  <div class="top">
+    <h1>Get Started</h1>
+    <div class="cta-copy">
+      <p>Build your first Ember application with our 5 minute tutorial.</p>
+    </div>
 
-  <div class="info">
-    <a href="/builds">More downloads</a>
+    <a href="https://guides.emberjs.com/v2.5.0/getting-started/quick-start/">View Quick Start guide</a>
```
rwjblue: This should use the same logic as the guides layout to figure out the version from the JSON file (IMO).

rwjblue:
One nitpick but this looks great!

source/about.html.erb
locks: 🤔 `versions.json` is in another domain, where do we configure `guides.emberjs.com`'s CORS?

rwjblue: It is done in Ruby land at build time, CORS isn't a factor.

PR #2566, opened 2016-04-25
Add Salsify to Ember users page

mattcross:
Salsify uses Ember for all of our customer-facing web applications and several internal web applications.


wifelette:
Looks like this needs a rebase, should be good to merge after :)

mattcross:
All set I think!

rwjblue:
Thanks!

PR #2571, opened 2016-04-28
emberjs/ember.js#10844 -- Update example usage of RSVP.Promise

menewman:
emberjs/ember.js#10844

### DIFF
--- a/data/api.yml
+++ b/data/api.yml
@@ -2382,7 +2382,9 @@ classes:
       ------------
 
       ```js
-      var promise = new Promise(function(resolve, reject) {
+      import RSVP from ‘rsvp’;
+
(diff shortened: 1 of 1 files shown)


rwjblue:
This file is an autogenerated compilation done by the YUIDoc tool in various repositories.  If we update this yaml file here, it will get overwritten by future updates.  We need to make the needed changes upstream in their respective repositories (and then update Ember to use those newer versions) for the changes to take effect.

Looks like all of these updates are for RSVP, would you mind submitting a pull request to that repo directly (here https://github.com/tildeio/rsvp.js/)?

menewman:
Oops, I've never worked with YUIDoc before. Thanks, I'll create a pull request for the RSVP repo.

PR #2572, opened 2016-04-28
Makes sponsors page responsive

locks:
http://quick.as/83kfOJ83


mixonic:
Seems like a good start :-)

PR #2593, opened 2016-05-20
Adds Travis CI continuous deployment

locks:
I can expand on the details tomorrow.


mixonic:
@locks awesome! Is there a TODO list or something to get this across the finish line?

locks:
@mixonic pressing the merge button, I think ;)
The thing I was going to expand on was the pipeline details, if anyone wants to know.

It's a rather straightforward script. I'm cloning and moving the files to it because we didn't want to force push to the deployed website. That way you can quickly pop in a revert, or something similar.
The committer is configured as Travis CI, if/when we get our own homu-like we can configure that.

This script will be expanded in the near future to pull in the extracted `builds` Ember CLI application.

PR #2595, opened 2016-05-21
responsive blog with pagination

webofnithin:
Added pagination to blog pages and made it responsive ! please close this #2594 duplicate Pull request

PR #2596, opened 2016-05-23
Tweak tags for blog posts

mixonic:
Blog post tagging had gotten kind of gnarly of late. For one, there was a "Recent Posts" tag that was applied in a varied number of ways, and additionally it was unclear how releases and ember-data releases should be tagged. Consequently there was a variety of junk on the left side of the page.

To improve the meaning of the categories we have, I've edited tags with these goals:

* An Ember.js release post should be tagged "Releases". It should not be tagged "Recent Posts". It may be tagged "Security".
* An Ember-Data release post should be tagged "Ember-Data". It should not be tagged "Recent Posts" or "Releases". It may also be tagged "Security"
* A non-release post, or a non-Ember non-Ember-Data release post, should be tagged "Recent Posts". Other tags are welcome.

There remain a series of old core team notes on the blog that I have not re-tagged to "Recent Posts". They seem ok on their own, even though there is no UI.

I've tweaked the count of displayed recent posts to be much shorter, as they are not super common.

Looks like:

screen shot 2016-05-23 at 11 22 19 am

/cc @bmac @tomdale @locks @acorncom @trek @wifelette


bmac:
lgtm 🏣

acorncom:
LGTM :shipit: I believe if we hit merge this should immediately show up on the site as well ...

acorncom:
Actually, @mixonic would the work that is going on in https://github.com/emberjs/website/pull/2595 alleviate some of the "Recent Posts" items you were trying to clean up here?  Not sure we should be removing the "Recent Posts" tag from older items ...

mixonic:
@acorncom that PR adds responsiveness and pagination, but I don't believe it addresses some of the things that editing the tags improves. For example in the screenshots he uses "Ember Data 0.13" is still a prominent link, obviously not so important we want it shown to visitors on the landing page.

I'm going to land this as-is, but let me know if there is more feedback!

mixonic:
Hm does not seem to have auto-deployed. Going to publish a build by hand.

PR #2600, opened 2016-05-25
Add Enumerable.contains to deprecation guide

bmeurant:
Deprecation guide update related to emberjs/ember.js#13553

[rendered](https://github.com/bmeurant/website/blob/master/source/deprecations/v2.x.html.md#enumerablecontains)

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -654,3 +654,74 @@ export default Ember.Component.extend({
 If you're an addon maintainer, there is a polyfill for safe string detection ([ember-string-ishtmlsafe-polyfill](https://github.com/workmanw/ember-string-ishtmlsafe-polyfill))
 that will help maintain backwards compatibility. Additionally, it's worth noting that `Ember.String.htmlSafe`
 is supported back to pre-1.0, so there should be no concerns of backwards compatibility there.
+
+#### Enumerable#contains
+
+##### until: 3.0.0
+##### id: ember-runtime.enumerable-contains
+
+The `Enumerable#contains` and `Array#contains` methods were deprecated in favor of `Enumerable#includes` and `Array#includes` 
+to stay in line with ES standards. See [RFC](https://github.com/emberjs/rfcs/blob/master/text/0136-contains-to-includes.md) for details.
+
+`contains` and `includes` have similar behaviors. A notable exception is how `NaN` values are handled.
+`contains` uses [Strict equality comparison algorithm](https://tc39.github.io/ecma262/2016/#sec-strict-equality-comparison)
+for testing inclusion while `includes` uses [SameValueZero algorithm](https://tc39.github.io/ecma262/2016/#sec-samevaluezero).
+
+Before:
+
+```js
+var arr = ['a', 'b', 'c', NaN, undefined, null];
+arr.contains('b');        // true
+arr.contains('d');        // false
+arr.contains(NaN);        // false
+arr.contains(null);       // false
+arr.contains(undefined);  // false
+```
+
+After:
+
+```js
+var arr = ['a', 'b', 'c', NaN, undefined, null];
+arr.includes('b');        // true
+arr.includes('d');        // false
+arr.includes(NaN);        // true
+arr.contains(null);       // true
+arr.contains(undefined);  // true
+```
+
+`includes` also allows a second optional parameter `startAt` to specify the index at which to begin searching:
+
+```js
+var arr = ['a', 'b', 'c', NaN];
+arr.includes('c', 2);   // true
+arr.includes('c', -2);  // true
+```
+
+Note that the second `startAt` parameter is only available for `Ember.Array` because `Ember.Enumerable` does not rely on index-ordered access.
+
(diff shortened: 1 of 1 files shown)


bmeurant:
No idea why Travis is complaining ... :-(

source/deprecations/v2.x.html.md
```diff
@@ -615,3 +615,45 @@ export default Router.map(function() {
   });
 });
 ```
+
+#### Enumerable.contains
+
+##### until: 3.0.0
+##### id: ember-runtime.enumerable-contains
+
+The `Enumerable.contains` and `Array.contains` methods were deprecated in favor of `Enumerable.includes` and `Array.includes`
```
mixonic: Can we represent theses as `Enumerable#contains`? I tend to treat `Enumerable.contains` as a method that is literally on that thing (class usually), and `#contains` as a method that exists on an instance of that thing (object, mixin applied to object).

source/deprecations/v2.x.html.md
```diff
@@ -615,3 +615,45 @@ export default Router.map(function() {
...
+
+#### Enumerable.contains
+
+##### until: 3.0.0
+##### id: ember-runtime.enumerable-contains
+
+The `Enumerable.contains` and `Array.contains` methods were deprecated in favor of `Enumerable.includes` and `Array.includes` 
+to stay in line with ES standards. See [RFC](https://github.com/emberjs/rfcs/blob/master/text/0136-contains-to-includes.md) for details.
+
+`contains` and `includes` have similar behaviours, excepting regarding `NaN` values management because includes uses
```
mixonic: Perhaps:

> `contains` and `includes` have similar behaviors. A notable exception is how `NaN` values are handled. `contains` uses [strict equality comparison](https://tc39.github.io/ecma262/2016/#sec-strict-equality-comparison) for testing inclusion while `includes` uses the [SameValueZero algorithm](https://tc39.github.io/ecma262/2016/#sec-samevaluezero).

* US-centric "behavior" 😬 
* Simplify the sentences a bit
* First describe the old behavior then the new

source/deprecations/v2.x.html.md
```diff
@@ -615,3 +615,45 @@ export default Router.map(function() {
...
+After:
+
+```js
+var arr = ['a', 'b', 'c', NaN];
+arr.includes('b');  // true
+arr.includes('d');  // false
+arr.includes(NaN);  // true
+```
+
+`includes` also allows a second optional parameter `startAt` to specify the index at which to begin searching for:
```
mixonic: > `includes` also allows a second optional parameter `startAt` to specify the index at which to begin searching:

mixonic:
I think this should also mention the behavior change of `addObject` and `without` to use `includes` instead of `contains`. It seems a pretty innocent breaking change, but it should be called out.

source/deprecations/v2.x.html.md
bmeurant: Thank you for correcting my english :-(

source/deprecations/v2.x.html.md
```diff
@@ -615,3 +615,74 @@ export default Router.map(function() {
...
+
+```js
+var arr = ['a', 'b', 'c', NaN];
+arr.includes('c', 2);   // true
+arr.includes('c', -2);  // true
+```
+
+Note that the second `startAt` parameter is only available for `Ember.Array` because `Ember.Enumerable` does not rely on index-ordered access.
+
+`Enumerable#without` and `MutableEnumerable#addObject` use now internally `includes` instead of `contains`. This leads to some changes:
```
mixonic: "some minor breaking changes" perhaps

mixonic:
@bmeurant this is 100% gtg with a small patch and a rebase. Sorry for the trouble, but you can land that I'd I will merge? :-D

bmeurant:
@mixonic sorry for the delay. I made the change and rebased. Should be ok now.

mixonic:
👏 thank you @bmeurant! :-D

source/deprecations/v2.x.html.md
```diff
@@ -654,3 +654,74 @@ export default Ember.Component.extend({
...
+```
+
+After:
+
+```js
+var arr = ['a', 'b', 'c', NaN, undefined, null];
+arr.includes('b');        // true
+arr.includes('d');        // false
+arr.includes(NaN);        // true
+arr.contains(null);       // true
```
mixonic: this line and the one below should be `includes`?

bmeurant: Ooooooops !  Yes it should be `includes`, I missed it, sorry. I can update but this is already merged :weary:

bmeurant: @mixonic let me know if you want me to make a new PR to fix this

locks: @bmeurant went ahead and [fixed it](https://github.com/emberjs/website/pull/2650) since the change is already deployed. Thanks for the original work :)

bmeurant: @locks :+1: thx. You're welcome

PR #2602, opened 2016-05-25
Ember Data 2.6 release post

bmac:


source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+  when just including the id of the records is not enough.
+
+Thanks to [@cibernox](https://github.com/cibernox) for
+implementing this feature.
+
+### Upcoming Features in Ember Data 2.7.beta-1
+
+#### `ds-boolean-transform-allow-null` [#4022](https://github.com/emberjs/data/pull/4022)
+
+Allow null/undefined values for boolean attributes via `DS.attr('boolean', { allowNull: true })`
```
locks: shouldn't null/undefined be backticked?

locks: new blueprints use the new modules, so it would be `attr`

source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+
+```js
+// app/models/user.js
+
+export default DS.Model.extend({
```
locks: same as above. how do we deal with this?

locks: also this example doesn't have an import and the RESTAdapter one does

source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+    ]
+  }
+}
+```
+
+This is particularly useful for polymorphic relationships not backed
+  by [STI](https://en.wikipedia.org/wiki/Single_Table_Inheritance)
+  when just including the id of the records is not enough.
+
+Thanks to [@cibernox](https://github.com/cibernox) for
```
acorncom: Do we have this documented in either the API docs or guides?

source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+This feature adds new public methods to `DS.RESTAdapter`, which allow to
+customize the properties of a request:
+
+- `methodForRequest` to get the HTTP verb
+- `urlForRequest` to get the URL
+- `headersForRequest` to get the headers
+- `dataForRequest` to get the data (query params or request body)
+
+The `params` hash passed to those methods has all the properties with
+which the corresponding `find`, `createRecord`, `findQuery`, ...  call
```
acorncom: ... _calls_

source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+have been invoked: store, type, snapshot(s), id(s) and query. The
```
acorncom: Backticks on these variables?

source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+
+Say your API handles creation of resources via PUT, this can now be
+customized as follows:
+
+``` js
+// adapters/application.js
+import DS from 'ember-data';
+
+export default DS.RESTAdapter.extend({
+  methodForRequest: function(params) {
```
acorncom: Should this be defined as an ES6 method?

`methodForRequest(params) {`

source/blog/2016-05-25-ember-data-2-6-released.md
```diff
@@ -0,0 +1,129 @@
...
+      return "PUT";
+    }
+
+    return this._super(...arguments);
+  }
+});
+```
+
+Thanks to [@pangratz](https://github.com/pangratz) for
+implementing this feature.
```
acorncom: It'd be great to get all this info in the guides on the adapters page as well ...  Seems like we could almost lift this verbatim to make it easier for folks to find instead of looking on the blog?

PR #2603, opened 2016-05-26
Minor CSS enhancements

zeppelin:
The original PR was made at the EmberConf Contributors Workshop. I've extracted these few commits (plus rebased onto master, of course), to be able to review & merge them in independently from other changes, such as the responsive stuff, etc.

PR #2604, opened 2016-05-27
Minor css enhancements reloaded

zeppelin:
Includes `@import "variables";` statements to fix `middleman build`, otherwise the same as #2603


acorncom:
Pulled this down to check locally and merge it in while resolving merge conflicts.  Thanks!

PR #2607, opened 2016-05-28
Fix travis builds so PR branches can have builds turned on again

acorncom:
More details on this can be found here: https://github.com/emberjs/website/pull/2603#issuecomment-221954665


acorncom:
@locks r+ ?

PR #2608, opened 2016-05-30
Fix broken meetups map

simonihmig:
The map on the meetups page (http://emberjs.com/community/meetups/) is broken because the linked markerclusterer script returns a 404. The google code repo is shut down, in favor of github. This PR fixes the URL.


locks:
Thanks for the fix!

PR #2622, opened 2016-06-09
Updated tag for Ember Data info

acorncom:
Matches the revisions that got made so that Ember Data releases are categorized separately from Ember releases on our sidebar ...


bmac:
Thanks @acorncom.

PR #2625, opened 2016-06-19
Updated the deprecation for `ember-htmlbars.ember-handlebars-safestring`

workmanw:
This makes mention of `Ember.String.isHtmlSafe` and the related polyfill.

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -591,7 +591,7 @@ export default Router.map(function() {
 ##### until: 3.0.0
 ##### id: ember-htmlbars.ember-handlebars-safestring
 
-Before:
+Creating safe strings. Before:
 
 ```js
 import Ember from 'ember';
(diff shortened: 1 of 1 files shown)


source/deprecations/v2.x.html.md
```diff
@@ -609,9 +609,48 @@ After:
...
+      if (Ember.String.isHtmlSafe(myString)) {
+        // ...
+      }
+    }
+  }
+});
+```
+
+If you're an addon maintainer, there is a polyfill for safe string detection ([ember-string-ishtmlsafe-polyfill](https://github.com/workmanw/ember-string-ishtmlsafe-polyfill))
+that will help maintain backwards compatibility. Additionally, it's worth noting that `Ember.Handlebars.htmlSafe`
```
rwjblue: I believe this should be `Ember.String.htmlSafe` (not `Ember.Handlebars.htmlSafe`).

rwjblue:
One small typo, but LGTM otherwise.  Thanks for tackling!

workmanw:
@rwjblue Once again, good catch! Fixed.

rwjblue:
Thanks!

Issue #2615, opened 2016-06-06
Deprecation Guide for modifying property within didInsertElement


Gaurav0:
The solution presented here:

http://emberjs.com/deprecations/v1.x/#toc_modifying-a-property-within-code-didinsertelement-code

usually doesn't work. didRender triggers the same deprecation warnings. In fact, this deprecation warning is poorly worded in that regard.

Anyway, I suggest that the best solution is to use the run loop, i.e. `Ember.run.schedule('afterRender', ...` within `didInsertElement`.

btecu:
This can be closed.

acorncom:
Thanks

PR #2631, opened 2016-06-26
Update suggested fix for deprecation warning

backspace:
As suggested by @Gaurav0 in #2615.

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -996,13 +996,15 @@ totalCount: Ember.computed('values.[]', function() {
 Because of changes in the way Ember renders components in 1.13, setting properties within an overridden `didInsertElement` method
 will result in a deprecation warning.
 
-In many cases you can move your logic later in the component lifecycle by implementing the `didRender` hook, on of
-[the new hooks introduced in 1.13](http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_component-lifecycle-hooks).
+You can wrap your changes in an `afterRender` run loop to silence these warnings:
 
 ```js
-  didRender() {
+  didInsertElement() {
     this._super(...arguments);
-    this.set('myValue', value);
+
+    Ember.run.schedule('afterRender', function() {
+      this.set('myValue', value);
+    });
   }
 ```
 


backspace:
During the LIVE ACTION PR CODE REVIEW SESSION, the result of some discussion was that there should be clarification here on how to fix this in a more desirable way, by rewriting your code to not produce double-renders. This suggestion of how to work around the warning if you _really_ need to double-render can remain, but should be marked as a last resort.

I’ll return to this soon!

locks:
I think of the proposed solution as more of an anti-pattern than a pattern. I'm closing this one, feel free to open another one when you return to the subject :)

Gaurav0:
@locks It would have been helpful if an acceptable solution is posted on the issue. "Rewrite your code so that no double render is necessary" is not a mechanical operation, and IMO means the deprecation itself should be reversed unless there is a way to fix it mechanically.

rwjblue:
@Gaurav0 - Nothing about the concept of a deprecation infers or suggests that they are all simple fixes. Many are complicated and involve programming model shifts (moving away from views in general is like this).

Gaurav0:
@rwjblue Not what was said: https://github.com/emberjs/rfcs/pull/56/files




-------- Original message --------

mixonic:
I'm not sure what is happening here.

* The deprecation guide here is incorrect. It needs updating.
* The author proposed a change and has received feedback, will follow up.
* The issue was closed and I'm unsure of the rational- this is just pending new work. I'll chat with @locks and see what was an anti pattern about the proposed route forward.
* The thread is being hijacked over some suggestion the deprecation be removed? Seems unrelated.

Please try to keep our scope on improvements to the documentation here. It is really needed, and I don't want to lose that end goal. Thanks!

Gaurav0:
@mixonic If there is no way to avoid a deprecation warning that we don't consider an antipattern, why do we have it deprecated? Why can we not correct an error on the website with a solution that works when there is no solution you guys like?

ef4:
@Gaurav0 there is no requirement that clearing deprecations is mechanical. That's why they are deprecations and not breaks. It's expected that sometimes they require real work, which is why we keep deprecated code around so long.

People who are upset by the message but don't have the time to fix it right away should be pointed toward ember-cli-deprecation-workflow.

Using your own run loop doesn't solve the problem, it just hides it, so it is equivalent to ember-cli-deprecation-workflow anyway. The real problem here is expensive render loops that are hurting app performance and causing data flow loop bugs.

Gaurav0:
@ef4 @mixonic 

In my experience, if you can't move the offending `set` to init because you are setting a value based on information you only have after render, there is no clean way to rewrite your code to avoid a double render. There really is no solution other than the one I suggested, which is why I think this deprecation should never have been made.

This deprecation message is extremely confusing to newcomers to Ember as I experienced this morning. Furthermore, it happens extremely often. I am really upset that we can't fix the guides to at least give some way out that works.

ef4:
We didn't add that message to make things harder, we did it to make things easier. Without the messages, all those newbies are just shooting themselves in the feet, because they will wonder why their app is so buggy and have no way of figuring out what's going on. 

It's a question of complaining early so the problem can be found, instead of failing silently. Those are the options. Make it magically work is not actually an option. It never worked reliably, because it can't without nuking performance for everyone.

I absolutely agree we need better teaching for how to react to the warning. The guide could walk through multiple scenarios and how to fix each one. Using a separate render loop _is_ one of those solutions but only in rare cases where you are truly measuring the output of the previous render. Newbies almost always have simple data flow problems that would be made worse by an extra run loop, not better.

Gaurav0:
> The guide could walk through multiple scenarios and fix each one.

Could someone describe a single nontrivial scenario where one could rewrite code to avoid a double render? (If the `set` can be cut and pasted to `init` or `didReceiveAttrs` that I would consider trivial.)

ef4:
What does trivial have to do with it? I already agreed it is non-trivial. If there was a trivial transformation we would just do that automatically.

Gaurav0:
You wanted me to walk through scenarios. I'm trying to figure out if there are any besides the obvious ones.

ef4:
Sorry, I misread your comment. :-p

ef4:
One of the common cases I see is a computed property getter that calls `set`. You can always replace the value being set with a separate computed property that depends on the one that used to be doing the `set`.

Another case is when there is a list of items and a currently highlighted item, and the user tries to store `isHighlighted` as a property on each item in the list, forcing them to try to make that data flow upward again through the parent. The fix is to move that state to the parent in the first place by storing a `highlightedItem` property there that always flows down to the children.

Gaurav0:
@ef4 Thanks. I'll put up another PR for this.

PR #2634, opened 2016-06-29
Update the website for the Ember Data 2.6.1 release

bmac:
Update the builds page for the Ember Data 2.7.0.beta.2 release


acorncom:
@bmac Would you mind sending in an update to https://github.com/ember-learn/builds as well?  We're trying to keep things in sync and will eventually be switching over to that app ...

PR #2635, opened 2016-06-30
Bug fix for quick start guide link on builds page

welzie:
Quick start guide was linking to versions that do not exist such as 2.6.1.  Now will only create links for latest MINOR versions such as 2.6.0.


locks:
Much appreciated :)

PR #2636, opened 2016-06-30
Fixed description of `ember.computed.and` method

tradziej:


locks:
Hi @tradziej, thanks for bringing this to our attention. This file is compiled from the [source code documentation](https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/computed/computed_macros.js), so the PR should be send in that repository instead :)

PR #2639, opened 2016-07-05
Fix minor display bug for users page

JohnONolan:
It's only a little thing on the [users](http://emberjs.com/ember-users/) page, but smaller logos should be center aligned to match the layout of the rest of the page:

**Before:**
image

**After:**
image


rwjblue:
Thank you @JohnONolan!

PR #2659, opened 2016-07-26
Fix alignment of sponsors columns + column contents

stefanpenner:
before:

screen shot 2016-07-26 at 11 16 02 am

after:

screen shot 2016-07-26 at 11 16 12 am

### DIFF
--- a/source/stylesheets/site.css.scss
+++ b/source/stylesheets/site.css.scss
@@ -1579,33 +1579,18 @@ ul.showcase {
 
   width: 33.333%;
   margin: 10px 0;
-  &:nth-child(3n+1) {
-    a {
-      margin: 0 auto 0 0;
-    }
-  }
-  &:nth-child(3n+2) {
-    a {
-      margin: 0 auto 0 auto;
-    }
-  }
-  &:nth-child(3n+3) {
-    a {
-      margin: 0 0 0 auto;
-    }
-  }
+
   a {
     background-image: url('/images/about/user_bg.png');
     display: block;
     height: 103px;
     width: 210px;
     text-align: center;
+    margin: 0 auto;
 
     img {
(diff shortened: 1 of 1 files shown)


stefanpenner:
`/sponsors` + `/ember-users` both appear ok

stefanpenner:
The existing CSS appears to have existed for a reason, but does not appear to work as originally intended. Removing it appears to have resulted in a fix, but no obvious regression. If there is something I'm missing, please share what should happen and I can look at as well.

chancancode:
@stefanpenner I think the goal is to use the full width of the page, i.e. have the left side of the rectangle aligned with the left side of the header/text.

source/stylesheets/site.css.scss
```diff
@@ -1579,33 +1579,18 @@ ul.showcase {
 
   width: 33.333%;
   margin: 10px 0;
-  &:nth-child(3n+1) {
-    a {
-      margin: 0 auto 0 0;
-    }
-  }
```
chancancode: Changing these to:

```css
&:nth-child(3n+1) {
  > * {
    display: block;
    margin: 0 auto 0 0;
    width: 210px;
  }		
}
```

...and so on, seems to more or less satisfy that goal

chancancode: result

chancancode: If you want to go with this, you would also need to target `> *` instead of `a` for other screen sizes

stefanpenner: If it's up to me, I'm ok with losing left-hand and right-hand margin alignment, as it is still pleasantly centered. It is possible to accomplish the alignment, but Im not terribly motivated to, so if someone is...

PR #2668, opened 2016-08-23
Added 'display: -webkit-flex' to support safari browsers < 9

Ajeey:
Browser version - Safari Version 8.0 (10600.1.25)

Before - 
screen shot 2016-08-22 at 12 40 58 pm

After - 
screen shot 2016-08-23 at 3 30 01 pm


locks:
Thanks for your contribution!

@homu r+

PR #2669, opened 2016-08-25
Updated Bidvine logo on Who's Using Ember page

slannigan:


wifelette:
Thanks @slannigan :) We have Daniel Stuart down as our contact for this if we have any logo questions in the future, please email listing@emberjs if that needs to be updated.

PR #2672, opened 2016-08-27
Clarify baseURL blog post.

nathanhammond:
Closes https://github.com/ember-cli/ember-cli/issues/6212


nathanhammond:
Rendered:
screen shot 2016-08-26 at 8 05 31 pm

sandstrom:
This is great! Essentially the same thought that I had (around the naming and `rootURL` being different things in different contexts). Is this something that should be added to the API docs too? 

Blog posts are great to highlight changes, but people rarely look in old blogposts when looking up API behaviour.

typeoneerror:
This is working for me for the most part, but with routerRootURL set, my api proxies are now also prepended with "/app/". Leaving the deprecated baseURL and it works like it used to. Is there another location I need to change to not affect the proxy?

PR #2679, opened 2016-09-09
Add link to Enumerable.includes polyfill.

blimmer:
It is helpful to have this information in the deprecation guide.


rwjblue:
Awesome! Would you mind linking from the 2.8 blog post also?

PR #2685, opened 2016-09-22
Add FreshBooks to the Using EmberJs List.

markstory:
The new FreshBooks interface is built with EmberJs.


data/users.yml
```diff
@@ -851,6 +851,10 @@
   image: wnyc.png
   alt: WNYC
   use: "WNYC's front-end is fully powered by Ember" 
+- url: https://freshbooks.com
+  image: freshbooks.png
+  alt: FreshBooks
+  use: "FreshBooks front-end application is powered by Ember"
```
aaronmw: Needs an apostrophe, no? `FreshBooks' front-end app...`

wifelette: Agreed, though I won't hold up the PR for that ;) I can add post-merge.

markstory: Thanks :+1:

PR #2687, opened 2016-09-30
Added video recording instruction section, some section reordering, Zoey fixes

wifelette:
The video stuff has come up a bunch of times from other newb meetup folks like myself, who would like to record their talks but have no experience in the area. Once someone asks more than twice, I say we add it to the page ;)

Thanks to @jgwhite for the initial equipment recommendations!

/cc @locks


acorncom:
👍

PR #2697, opened 2016-10-12
replace npm with yarn in gett started

### LINKED ISSUE #2696: Update Getting Started on homepage to use yarn
`yarn global add ember-cli` instead of `npm install -g ember-cli` [docs](https://yarnpkg.com/en/docs/migrating-from-npm). 

Or maybe give both options?

screen shot 2016-10-12 at 12 48 43 pm

iezer:
Fixes #2696 
screen shot 2016-10-12 at 12 48 43 pm

### DIFF
--- a/source/about.html.erb
+++ b/source/about.html.erb
@@ -40,7 +40,7 @@ title: "A framework for creating ambitious web applications."
     <a href="https://guides.emberjs.com/<%= current_version %>/getting-started/quick-start/">View Quick Start guide</a>
   </div>
   <div class="bottom">
-    <p>$ npm install -g ember-cli</p>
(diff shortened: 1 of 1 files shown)


stefanpenner:
@iezer we are totally pushing towards this being possible. Currently the known blocker is: https://github.com/yarnpkg/rfcs/pull/1

rwjblue:
I suspect that we have to support both things either way (even if we internally use and suggest Yarn, I can't imagine ember-cli deciding to not support `npm` also)

stefanpenner:
related issue on CLI: https://github.com/ember-cli/ember-cli/issues/6341

PR #2699, opened 2016-10-14
Tomster filter

ultimatemonty:
This popped up this evening when I was showing my daughter all the different Zoeys and I thought "man it'd be nice to be able to see just the Zoeys".

Let there be filtering!

Caveat: I know **nothing** about Ruby/Middleman and figured this out mostly by Googling so I likely did some ignorant things. Please point these things out so I can make them mo betta!

### DIFF
--- a/source/tomster.html.erb
+++ b/source/tomster.html.erb
@@ -2,6 +2,10 @@
 title: "Tomster and Zoey"
 ---
 
+<% content_for :foot do %>
+  <%= javascript_include_tag 'tomster' %>
+<% end %>
+
 <div class="tomster section intro">
   <h1>Tomster and Zoey</h1>
   <p>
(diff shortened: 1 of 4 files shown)


wifelette:
Pulled this down and took a look, this is great! 

I can't decide if I think the category titles should be pluralized, a la Tomsters/Zoeys/Meetups/Conferences/etc. But I _think_ I think that? Thoughts, @ultimatemonty?

ultimatemonty:
@wifelette I'm good either way :)

Did the ruby/Middleman side of things looks good? One thing that was kind of bothering me was the dynamic URLs end in a `/` and it seems like they shouldn't. Probably an easy way to fix that but I wasn't able to find anything on the Googles

Also, sorry for dropping a surprise PR on you! I realized after the fact (and @acorncom kindly pointed it out on Slack) that opening an issue before PR'ing is the right approach for this kind of stuff. Will make sure to do so in the future!

wifelette:
Having slept on it, let's pluralize :p Can you make the tweak?

RE: Middleman/Ruby: truthfully while I use it a lot, my level of expertise isn't really... existing ;) I've learned to do the very basics of what I need and not much more than that.

Indeed it could probably be prettier—your work _and_ the work done prior—but lacking an available expert I say we ship it as is and maybe someone else can tidy it later if indeed there's a better way. Or if anyone reading this _is_ said expert, by all means, please jump in!

ultimatemonty:
@wifelette will do later this evening!

source/tomster.html.erb
```diff
@@ -1,20 +1,29 @@
 ---
 title: "Tomster and Zoey"
+filters: all, tomster, zoey
 ---
 
 <div class="tomster section intro">
   <h1>Tomster and Zoey</h1>
   <p>
-    Tomster was the first of many friendly faces of the Ember project and community. Over the years we've dressed and redressed him for all sorts of fun projects and reasons. With the addition of his sister Zoey, you'll see Ember friends around even more.
+    Tomster was the first of many friendly faces of the Ember project and community. Over the years we"ve dressed and redressed him for all sorts of fun projects and reasons. With the addition of his sister Zoey, you"ll see Ember friends around even more.
```
acorncom: The quote here changed from a single to a double quote, could you change that back?

source/tomster.html.erb
```diff
@@ -1,20 +1,29 @@
...
 </div>
 
 <div class="tomster section list">
+  <div id="tomster-filter">
+    <%= active_link_to "All", "/tomster/" %> |
+    <%= active_link_to "Tomster", "/tomster/tomster/" %> |
+    <%= active_link_to "Zoey", "/tomster/zoey/" %> |
+    <%= active_link_to "Meetup", "/tomster/meetup/" %> |
+    <%= active_link_to "Conference", "/tomster/conference/" %> |
+    <%= active_link_to "Other", "/tomster/other/" %>
```
acorncom: Let's change the above links to `/tomster?filter=tomster` and equivalent  Doing that will make life simpler if/when we move this page to Ember (as we don't have to worry about having additional pages to map on the search engine side of things).

**Edit**: hmm, looks like that's not an option we can accommodate with Middleman, carry on ... ;-)

ultimatemonty:
@acorncom @locks @wifelette I believe I've addressed all of your feedback

* Filter now works via queryParams instead of dynamic pages (for future compat)
* Use flexbox for layout as opposed to tables
* Fixed copy pasta errors

ultimatemonty:
@locks commits squashed

locks:
Thank you for your contribution @ultimatemonty 😁  Keep'em coming!

acorncom:
Very snazzy, thanks @ultimatemonty 👍

PR #2700, opened 2016-10-15
Add deprecation guide for {{render helper

josemarluedke:
This adds the deprecation guide for emberjs/ember.js#14441.

This guide could be expanded and reworded with some feedbacks. 

@ErikCH has shown interest in taking this documentation, but I wanted to get at least a basic documentation for this deprecation. Anyone else with more experience writing deprecations and using the `{{render` helper is welcome to take this further.

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -753,3 +753,107 @@ component.appendTo('.my-component-wrapper');
 ```
 
 Note that both APIs are private, so no public API is been deprecated here.
+
+
+#### `{{render` helper
+
+##### until: 3.0.0
+##### id: ember-template-compiler.deprecate-render
+
+Using the `{{render` helper is deprecated in favor of using components.
+Please refactor uses of this helper to components:
+
+For example, if you had:
+
+```hbs
+{{render 'my-sidebar'}}
+```
+
+```hbs
+{{! app/templates/my-sidebar.hbs }}
+<p>template stuff here</p>
+```
+
+```js
+// app/controllers/my-sidebar.js
+export default Ember.Controller.extend({
+});
+```
+
+You would refactor to a component like so:
+
+```hbs
+{{my-sidebar}}
+```
+
+```hbs
+{{! app/templates/components/my-sidebar.hbs }}
+<p>template stuff here</p>
+```
+
+```js
+// app/components/my-sidebar.js
+export default Ember.Component.extend({
+});
+```
+
+Note that the render helper has several unique behaviors that may require further refactoring work during migration to a component.
+
+- When using the render helper with no model argument, the controller instance is a singleton. For example the same controller instance is shared between `{{render 'post'}}`, any other helper usage of `{{render 'post'}}`, a route template named post, and dependency injections using `Ember.inject.service('post')`.
+- When sendAction is called in a rendered controller, or when `{{action` is used in a render helper template, the bubbling target for those actions is the router and current active route. With components, those same actions would target only the component instance without bubbling.
+
+
+#### Rendering into a {{render}} helper that resolves to an {{outlet}}.
+
+##### until: 3.0.0
+##### id: ember-routing.top-level-render-helper
+
+Before named outlets were introduced to Ember the render helper was used to declare slots for `this.render` in routes. This usage is not common in modern, idiomatic applications and is deprecated. In general, the pattern of named outlets or named render helpers is discouraged. Instead use of [ember-elsewhere](https://github.com/ef4/ember-elsewhere) or another DOM-redirection library should better serve these use cases.
+
+For example this code uses the render helper as a target for a special sidebar present on the index route. The special sidebar is in a template named `index-sidebar`:
+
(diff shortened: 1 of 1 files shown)


josemarluedke:
I have added more guide for the deprecation of outlet orphaning (route `this.render({into: ...})`) requested here: https://github.com/emberjs/ember.js/pull/14441#issuecomment-253964838

josemarluedke:
@locks could you take a look on this one? Since emberjs/ember.js#14441 has been merged and the the release of Ember 2.10 is coming soon, I think we should have a look at merging it or doing the required changes.

source/deprecations/v2.x.html.md
```diff
@@ -753,3 +753,92 @@ component.appendTo('.my-component-wrapper');
...
 Note that both APIs are private, so no public API is been deprecated here.
+
+
+#### `{{render` helper
+
+##### until: 3.0.0
+##### id: ember-template-compiler.deprecate-render
+
+Using the `{{render` helper is deprecated in favor of using components.
+Please refactor to a component and invoke thusly:
```
mixonic: `and invoke thusly:` followed by another line with a `:` seems odd. Perhaps just "Please refactor uses of this helper to components".

source/deprecations/v2.x.html.md
```diff
@@ -753,3 +753,92 @@ component.appendTo('.my-component-wrapper');
...
+For example, if you had:
+
+```hbs
+{{! app/templates/application.hbs }}
+{{render "sidebar"}}
+```
+
+```hbs
+{{! app/templates/sidebar.hbs }}
+<div class="sidebar">{{outlet}}</div>'
```
mixonic: Extra `'` on this line

source/deprecations/v2.x.html.md
```diff
@@ -753,3 +753,92 @@ component.appendTo('.my-component-wrapper');
...
+```
+
+```js
+// app/controllers/my-sidebar.js
+App.IndexRoute = Ember.Route.extend({
+  renderTemplate() {
+    this.render({ into: 'sidebar' });
+  },
+  actions: {
+    disconnect: function() {
```
mixonic: shorthand function syntax was used for `renderTemplate`, the same should be used here. `disconnect() {`

source/deprecations/v2.x.html.md
```diff
@@ -753,3 +753,92 @@ component.appendTo('.my-component-wrapper');
...
+export default Ember.Component.extend({
+});
+```
+
+#### Rendering into a {{render}} helper that resolves to an {{outlet}}.
+
+##### until: 3.0.0
+##### id: ember-routing.top-level-render-helper
+
+For example, if you had:
```
mixonic: It would be nice to explain what the functionality is before giving an example. I expect most developers are not familiar with this functionality, and if you don't explain what it is you'll give them palpitations trying to guess if their codebase needs to change :-)

source/deprecations/v2.x.html.md
```diff
@@ -753,3 +753,92 @@ component.appendTo('.my-component-wrapper');
...
+```hbs
+{{! app/templates/components/my-sidebar.hbs }}
+<p>template stuff here</p>
+```
+
+```js
+// app/components/my-sidebar.js
+export default Ember.Component.extend({
+});
+```
```
mixonic: Can we list some common gotchas of making this migration? For example:

> Note that the render helper has several unique behaviors that may require further refactoring work during migration to a component.
> * When using the render helper with no `model` argument, the controller instance is a singleton. For example the same controller instance is shared between `{{render 'post'}}`, any other helper usage of `{{render 'post'}}`, a route template named `post`, and dependency injections using `Ember.inject.service('post')`.
> * When `sendAction` is called in a rendered controller, or when `{{action` is used in a render helper template, the bubbling target for those actions is the router and current active route. With components, those same actions would target only the component instance without bubbling.

What else?

mixonic:
@josemarluedke Thanks for bringing this up again. I've left a number of comments to tighten things. Do try to think of these docs from the perspective of a possibly overwhelmed developer who didn't keep track of the discussions we've had here on Github. You should provide some small amount of motivation for the change, liberally link back to discussions on GitHub you think are relevant, and try to provide as much actionable information as is possible.

source/deprecations/v2.x.html.md
mixonic: @josemarluedke I don't think anyone is very familiar with this API :-) Hence the strooong motivation to remove it :-)

I took a pass at an updated section [in this gist](https://gist.github.com/mixonic/b7a1051eaffa8db62027c4cfdbe1d2f0). Please give it a review, I may have missed something. It should be a good start to replace this section though, if you can merge that draft into this after review that seems good 👍

josemarluedke: That is perfect @mixonic! Thank you for getting that gist. I have applied your suggestions in this PR. Should be way better now.

mixonic:
@josemarluedke thanks :-D 👏

PR #2701, opened 2016-10-15
Twiddlefy Homepage Examples (WIP)

ultimatemonty:
First attempt at Twiddlefying the homepage examples. Working for the most part but there are a couple of issues that need to be sorted:
- Had to remove the `?fullScreen=true` queryParam from the embedded iframe URL due to some layout issues:
  image
- There are a few layout issues because the container for the homepage is smaller than Twiddle prefers (see header near **Dependencies** and the Build status in the **Output** pane)
  image

Any ideas on how to address these?

### DIFF
--- a/source/about.html.erb
+++ b/source/about.html.erb
@@ -62,50 +62,40 @@ title: "A framework for creating ambitious web applications."
 </div>
 
 <div class="getting-started section">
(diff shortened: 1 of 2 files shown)


locks:
We went another route, but thank you for your effort!

PR #2702, opened 2016-10-16
Add deprecation guide for renderToElement

josemarluedke:
Deprecation for emberjs/ember.js#14482.

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -729,3 +729,27 @@ export default Router.map(function() {
   });
 });
 ```
+
+### Deprecations Added in 2.10
+
+#### `renderToElement`
+
+##### until: 2.12.0
+##### id: ember-views.render-to-element
+
+Using the `renderToElement` is deprecated in favor of `appendTo`.
+Please refactor to use `appendTo`:
+
+For example, if you had:
+
+```js
+component.renderToElement('.my-component-wrapper');
+```
+
+Would be refactored to:
+
+```js
+component.appendTo('.my-component-wrapper');
+```
+
+Note that both APIs are private, so no public API is been deprecated here.


josemarluedke:
Note to self: We should explain that both functions are private and no public api is been deprecated here.

josemarluedke:
Added note about both api been private and updated the `until` to reflect the change in the Ember pr.

source/deprecations/v2.x.html.md
```diff
@@ -729,3 +729,27 @@ export default Router.map(function() {
   });
 });
 ```
+
+### Deprecations Added in 2.10
```
rwjblue: Can you tweak this to 2.11 (it didn't make the 2.10 cycle)?

rwjblue:
Landed for now (so the URL referenced in Ember is still present), but would still love a followup PR to tweak that version...

chancancode:
@josemarluedke Thanks for working on this!

I noticed a problem with the deprecation guide – I think the `renderToElement` usage here is incorrect. The method takes a tag name as argument (e.g. `"div"`), creates a new orphan element of that type then append the component to it. So I think the comparable replacement is to first create the element you wanted and then pass it to `appendTo`.

@rwjblue does that sound right to you?

josemarluedke:
@chancancode I got a PR opening a quick fix for it. Let me know if that change does make sense.

source/deprecations/v2.x.html.md
```diff
@@ -729,3 +729,27 @@ export default Router.map(function() {
...
+component.renderToElement('.my-component-wrapper');
+```
+
+Would be refactored to:
+
+```js
+component.appendTo('.my-component-wrapper');
+```
+
+Note that both APIs are private, so no public API is been deprecated here.
```
thewalkingtoast: `been` => `being`

locks: Can you send a PR for that change please?

thewalkingtoast: No

PR #2705, opened 2016-10-18
Fix solution for Modifying a Property within `didInsertElement`; walkthrough scenarios

Gaurav0:
Fix for #2615, incorporating feedback in #2631

### DIFF
--- a/source/deprecations/v1.x.html.md
+++ b/source/deprecations/v1.x.html.md
@@ -1006,7 +1006,7 @@ In many cases you can move your logic earlier in the component lifecycle by impl
   }
 ```
 
-It may even be possible to move the change to `init`, where the value can only be set once.
+It may even be possible, if your value is constant, to move the change to `init`, where the value can only be set once.
 
 ```js
   init() {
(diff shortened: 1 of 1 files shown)


source/deprecations/v1.x.html.md
```diff
@@ -996,16 +996,44 @@ totalCount: Ember.computed('values.[]', function() {
...
+It may even be possible to move the change to `init`, where the value can only be set once.
+
+```js
+  init() {
+    this._super(...arguments);
+    this.set('myValue', myValue);
+  }
+```
+
+Don't see a `set` in your `didInsertElement`? It may be in `didUpdate` or `didRender`. Due to a bug, these run during the same process.
```
rwjblue: The bug here is that they run and report the `didInsertElement` as the method causing the error.  They will (and should) be run roughly at the same time as `didInsertElement`, but the error/deprecation/whatever had the wrong method name.

source/deprecations/v1.x.html.md
```diff
@@ -996,16 +996,44 @@ totalCount: Ember.computed('values.[]', function() {
...
+
+Still can't find it? One of your computed properties may be calling `set`. Or one of your child components may be setting one of its attributes and it is being propagated upward to this component.
+
+In rare cases, you may be measuring the DOM rendered and want to set an attribute based on that. In this case, it is ok to cause a second render:
+
+```js
+  didInsertElement() {
+    this._super(...arguments);
+    run.scheduleOnce('afterRender', this, this.measureAndSet);
```
rwjblue: Since `didInsertElement` is only called once, I don't think `render.scheduleOnce` is needed (though it isn't bad, it just takes a small amount more work to dedupe calls than `run.schedule`).

Gaurav0: Fixed.

Gaurav0:
@rwjblue Responded.

mixonic:
Thanks @Gaurav0!

PR #2709, opened 2016-10-24
Update Billy image and url

jryanwestby:


wifelette:
@jryanwestby Thanks for getting this in! 

The old entry you're updating is from a time before we had the `use` field—you can see it on all the more recent entries. Can you add it to yours and then this should be good to merge?

locks:
Hey @jryanwestby, waiting on that `use` field so we can merge this :)

rwjblue:
@jryanwestby - Ping?

jryanwestby:
Sorry about the delay on this. I've added that use field.

wifelette:
So close ;) Shouldn't be a product description, but a description of the Ember use. A la "Billy's main app is built with Ember."

locks:
@jryanwestby nearly there, can you address @wifelette's latest comment so we get this puppy on the website? :]

jryanwestby:
There we go. Thank you for your patience and sorry about the delay

locks:
And we're off to the races!

PR #2711, opened 2016-10-26
Clarify escapeCSS use.

tabeth:
#escapeCSS is not natively implemented by Ember. Closes #14518.


locks:
Thanks!

QuantumInformation:
Would probably be nice to link to an example implementation.

PR #2714, opened 2016-11-03
Add indentation to fix line numbers on homepage example

nicomihalich:
Current website looks like (6 != 7) 👎 :
screen shot 2016-11-03 at 12 37 30 pm

But if we fix the indentation, it looks like (7 == 7) :+1: :

screen shot 2016-11-03 at 12 37 35 pm


rwjblue:
Thank you!

PR #2722, opened 2016-11-25
Add showcase applications to learn page

locks:
- [x] Builds app
- [x] Super Rentals

Fixes https://github.com/emberjs/website/issues/2749.


source/learn.html.erb
```diff
@@ -12,9 +12,9 @@ responsive: true
   <p>Are you getting started on your Ember.js adventure? Check out our official reading material: </p>
 
   <ul>
-    <li><a href="https://guides.emberjs.com/v2.9.0/getting-started/quick-start/"><strong>Quick start</strong> <i class="icon-link-ext"></i></a>: An easy breazy introduction to the framework.</li>
-    <li><a href="https://guides.emberjs.com/v2.5.0/tutorial/ember-cli/"><strong>Tutorial</strong> <i class="icon-link-ext"></i></a>: Follow our tutorial to build, and deploy, your first Ember.js application!</li>
-    <li><a href="https://guides.emberjs.com/"><strong>Guides</strong> <i class="icon-link-ext"></i></a>: Whether you're just getting started and want to get familiar with Ember, or you're looking to refresh your knowledge on a certain feature, the Guides are the place for you.</li>
+    <li><strong>Quick start</strong> <a href="https://guides.emberjs.com/v2.9.0/getting-started/quick-start/"><i class="icon-link-ext"></i></a>: An easy breazy introduction to the framework.</li>
```
acorncom: easy *breazy*? Would suggest dropping the *breazy* (and possibly the easy too?)

source/learn.html.erb
```diff
@@ -12,9 +12,9 @@ responsive: true
...
+    <li><strong>Tutorial</strong> <a href="https://guides.emberjs.com/v2.5.0/tutorial/ember-cli/"><i class="icon-link-ext"></i></a>: Follow our tutorial to build, and deploy, your first Ember.js application!</li>
+    <li><strong>Guides</strong> <a href="https://guides.emberjs.com/"><i class="icon-link-ext"></i></a>: Whether you're just getting started and want to get familiar with Ember, or you're looking to refresh your knowledge on a certain feature, the Guides are the place for you.</li>
```
acorncom: Seems like it'd be good to use the `current_version` helper for these links as well so we don't have to come back and update regularly ... (see it in action here: https://github.com/emberjs/website/blob/master/source/about.html.erb#L40)

source/learn.html.erb
```diff
@@ -53,5 +53,22 @@ responsive: true
...
+    <div class="showcase--item">
+      <h2>Super Rentals</h2>
+      <p>The outcome of our Tutorial</p>
+    </div>
+    <div class="showcase--item">
+      <h2>Builds</h2>
+    </div>
+    <div class="showcase--item">
+      <h2>???</h2>
+    </div>
```
acorncom: Help wanted here once it gets done.  http://ember-engines.com ?

locks:
@acorncom I believe I've addressed your concerns. We should keep an issue open with showcase suggestions!

locks:
Current look:
screen shot 2016-12-10 at 18 14 44

I need to figure out the screenshot of the application, but I'm ok with shipping this and iterating. cc @wifelette

wifelette:
I'd like to insist that the screenshots all be some specific size, for visual congruity. Like on the user page, we can note the exact required dimensions in the YAML file and only accept PRs that add the correctly-sized images. 

In my comment on the YAML file thread I had a description1 and description2 field, because that's what I think I see here, but I'm not 100% clear on why there are two separate bullets for each screenshot. Are there general guidelines for what the difference between the two lines might be?

locks:
screen shot 2016-12-11 at 10 12 09

Standardized image size.

locks:
Small tweak:
screen shot 2016-12-11 at 15 38 42

acorncom:
@locks LGTM :+1:

PR #2730, opened 2016-11-29
Fix deprecation example of renderToElement

josemarluedke:
It seems that the deprecation example was wrong. This updates it as per @chancancode's comment on https://github.com/emberjs/website/pull/2702#issuecomment-263493004.

Let me know if the change does make sense. Also should we mention that the `appendTo` can also receive an string with the selector?


chancancode:
👀 looks good to me, @rwjblue?

PR #2731, opened 2016-11-29
[BUGFIX] Builds sub-routes

### LINKED ISSUE #2725: Builds sub-routes result in 404
Middleman isn't properly configured, so for example http://emberjs.com/builds/release goes to 404.

Patsy-issa:
Fixes #2725


josemarluedke:
@Patsy-issa it seems that with this change, we are now getting 404 in http://emberjs.com/builds/.

PR #2736, opened 2016-12-01
Update 2.10 release post now that CLI beta is published.

nathanhammond:
That is all.


mixonic:
👏 Thank you @nathanhammond :-)

PR #2737, opened 2016-12-01
Fix linking to incorrect version of quick start guide

benjaminparnell:
Hey there :wave: 

In its current state the versions helper gives v2.9.0 as the current version, causing the quick start link on the homepage to link to the quick start guide for that version instead of v2.10.0.

This is because [this list](https://raw.githubusercontent.com/emberjs/guides.emberjs.com/master/snapshots/versions.json) that the versions are read from is not in order. I fixed this by sorting the list of versions using the Gem class.

### DIFF
--- a/lib/versions.rb
+++ b/lib/versions.rb
@@ -9,7 +9,9 @@ module Versions
   module Helpers
 
     def current_version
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you!

PR #2747, opened 2016-12-08
Add deprecation guide for queryRecord array responses

HeroicEric:
This adds the `ds.serializer.rest.queryRecord-array-response` deprecation to the deprecation guide for ember data.

### DIFF
--- a/source/deprecations/ember-data/v2.x.html.md
+++ b/source/deprecations/ember-data/v2.x.html.md
@@ -163,6 +163,99 @@ See [Disabling Prototype
 Extensions](https://guides.emberjs.com/v2.10.0/configuring-ember/disabling-prototype-extensions/#toc_strings)
 for more information about how Ember uses prototype extensions.
 
+#### Store.queryRecord Array Response with RESTSerializer
+
+###### until: 3.0.0
+###### id: ds.serializer.rest.queryRecord-array-response
+
+When using
+[`DS.RESTSerializer`](http://emberjs.com/api/data/classes/DS.RESTSerializer.html)
+with previous versions of Ember Data,
+[`store.queryRecord`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord)
+provided support for normalizing payloads containing an array of primary data.
+This behavior has been deprecated because it is basically the same as using
+[`store.query`](http://emberjs.com/api/data/classes/DS.Store.html#method_query)
+and returning the first model.
+
+Deprecated payload example with an array as the primary data:
+
+```js
+// GET /users?username="GummyBear"
+
+{
+  "users": [{
+    "id": "1",
+    "username": "GummyBear"
+  }]
+}
+```
+
+Expected payload example with a single object as the primary data:
+
+```js
+// GET /users?username="GummyBear"
+
+{
+  "user": {
+    "id": "1",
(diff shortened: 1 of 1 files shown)


HeroicEric:
@pangratz would you mind reviewing?

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -110,3 +110,85 @@ Note: The existing behavior of
...
+[`DS.RESTSerializer`](http://emberjs.com/api/data/classes/DS.RESTSerializer.html)
+with previous versions of Ember Data,
+[`store.queryRecord`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord)
+provided support for normalizing payloads containing an array of primary data.
+This behavior has been deprecated.
+
+Deprecated payload example with an array as the primary data:
+
+```js
+// GET /users?firstName="eric"
```
pangratz: I think an example using an actual unique key would be a little more helpful here. Maybe `username` instead? The reasoning would be that there are API use cases which always provide a single object and that is the main use case for `queryRecord`: query for a single record where the id is not known beforehand.

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -110,3 +110,85 @@ Note: The existing behavior of
...
+
+###### until: 3.0.0
+###### id: ds.serializer.rest.queryRecord-array-response
+
+When using
+[`DS.RESTSerializer`](http://emberjs.com/api/data/classes/DS.RESTSerializer.html)
+with previous versions of Ember Data,
+[`store.queryRecord`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord)
+provided support for normalizing payloads containing an array of primary data.
+This behavior has been deprecated.
```
pangratz: I think it would be good to point out why this has been deprecated: because it basically not different to a `store.query` with getting the first object of the response.

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -110,3 +110,85 @@ Note: The existing behavior of
...
+    "firstName": "Eric",
+    "lastName": "Kelly"
+  }
+}
+```
+
+If you need to support an API that responds with an array as the primary data,
+you can override
+[`normalizeQueryRecordResponse`](http://emberjs.com/api/data/classes/DS.RESTSerializer.html#method_normalizeQueryRecordResponse)
+in your serializer, manipulating the payload so it matches the expected format:
```
pangratz: I think we should point out the `store.query(…).then((result) => { result.objectAt(0) })` approach before, as this might be more straightforward for people to implement.

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -110,3 +110,85 @@ Note: The existing behavior of
...
+// app/adapters/user.js
+
+import DS from 'ember-data';
+
+export default DS.RESTAdapter.extend({
+  urlForQueryRecord() {
+    let baseURL = this.buildURL();
+    return `${baseURL}/user-query`;
+  }
+});
```
pangratz: 👍

HeroicEric:
thanks for the review @pangratz 😄

PR #2754, opened 2016-12-12
Add ids on FAQ section

chadhietala:
Would be nice to be able to link directly FAQs and items in the FAQs.


locks:
Was about to do this ❤️

PR #2763, opened 2016-12-15
Introduce a new Cookbook section

locks:
The suggestion for the first recipe is: "Integrating Ember in your backend framework".

We should provide instructions for at least two different frameworks, being the first two suggestions:
- [ ] Ruby on Rails
- [ ] Drupal

Help very much wanted, so leave a comment if you want to contribute either one of the listed ones, or for another framework.

### DIFF
--- a/source/cookbook.html.erb
+++ b/source/cookbook.html.erb
@@ -1,0 +1,16 @@
+---
+title: Cookbook
+responsive: true
+---
+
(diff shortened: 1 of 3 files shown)


jenweber:
Closed? I think if we wrote a cookbook, it'd be in the ember app anyway.

Issue #2761, opened 2016-12-15
Add deprecation guide for ember-views.lifecycle-hook-arguments


locks:
- RFC: https://github.com/emberjs/rfcs/pull/191
- Ember PR: https://github.com/emberjs/ember.js/pull/14711

As per the deprecation RFC process, it is necessary to write a corresponding [deprecation guide](http://emberjs.com/deprecations). While the RFC is not merged yet, meaning the deprecation is not in the codebase, much less part of a release, we can preemptively write the guide so everything is ready to go.

The steps should be roughly the following:
1. Read the RFC (link above), specifically "Transition Path" and "How We Teach This"
2. Add a [Deprecations Added in Pending Features](https://github.com/emberjs/website/blob/master/source/deprecations/ember-data/v2.x.html.md#deprecations-added-in-pending-features) header to the Ember 2.x file
3. Add a deprecation guide for `ember-views.lifecycle-hook-arguments`

We need step 2 since it'll be the first of its kind in the Ember 2.x file, and because the RFC still doesn't have a release as mentioned above.
For step 3 it needs to be the following structure:

 ```
#### <Title of the deprecation>

##### until: <Ember version where this might be removed>
##### id: <id of the deprecation>

<explanation of how to update the code in order to resolve the deprecation>
```

You can grab the id from the Ember PR linked at the top of the post.
The `until` version is usually 3.0.0, but since this is a deprecation of a private API it should be the version after the next LTS. You can also find this information in the [Ember PR](https://github.com/emberjs/ember.js/pull/14711/files#diff-7536fca9f7b528326424427194fd9ac2R511)!
The explanation should be a combination of "Transition Path" and "How We Teach This".

cc @jenweber

PR #2765, opened 2016-12-18
First draft of component hook argument deprecation guide

jenweber:
@locks here is the PR to address https://github.com/emberjs/website/issues/2761

I mostly used your examples from the RFC. I changed the "map" example to "thermometer" because when I read map, my mind jumps to enumerating through arrays. I don't think about navigational maps. I simplified what happens inside the component hooks. I see why there were more lines in terms of a real-world example but I think keeping it as simple as possible may help the reader to hone in on what's different between the before and after.

I am also wondering what the best heading level is for each hook. They don't stand out well but they also should not be as prominent as the label.

What's the official deprecation name? I made one up.

Thanks for walking me through the process!

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -679,57 +679,6 @@ to fix the deprecation in a backwards-compatible way.
 
 Added in [PR #13553](https://github.com/emberjs/ember.js/pull/13553).
 
-### Deprecations Added in Pending Features
-
-#### Route#serialize
-
-##### until: 3.0.0
-##### id: ember-routing.serialize-function
-
-The `Route#serialize` function was deprecated in favor of passing a `serialize` function into the route's definition in the router map. For more detailed information see the [Route Serializers RFC](https://github.com/emberjs/rfcs/blob/master/text/0120-route-serializers.md).
-
-As an example, given a route like:
-
-```js
-// app/routes/post.js
-import Ember from 'ember';
-export default Ember.Route.extend({
-  // ...
-  serialize(model) {
-    return { post_id: model.id };
-  }
-});
-
-// app/router.js
-export default Router.map(function() {
-  this.route('post', {
-    path: '/post/:post_id'
-  });
-});
-```
-
-You would refactor it like so:
-
-```js
-// app/routes/post.js
-import Ember from 'ember';
-export default Ember.Route.extend({
-  // ...
-});
-
-// app/router.js
-function serializePostRoute(model) {
-  return { post_id: model.id };
-}
-
-export default Router.map(function() {
-  this.route('post', {
-    path: '/post/:post_id',
-    serialize: serializePostRoute
-  });
-});
-```
-
 ### Deprecations Added in 2.11
 
 #### `renderToElement`


locks:
> What's the official deprecation name? I made one up.

You can see that Godfrey used `ember-views.lifecycle-hook-arguments` in his https://github.com/emberjs/ember.js/pull/14711/files#diff-7536fca9f7b528326424427194fd9ac2R514, so you got that right :)

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -517,3 +517,95 @@ let author = this.store.peekRecord('user', 456);
 
 post.set('author', author);
 ```
+
+#### Arguments in Component Lifecycle Hooks
+
+##### until: 2.13.0
+##### id: ember-views.lifecycle-hook-arguments
+
+It is possible for component lifecycle hooks `didInitAttrs`, `didReceiveAttrs`, and `didUpdateAttrs` to receive arguments, but this functionality is part of private API, which is subject to change without notice. Those arguments will be removed because their use is harmful to component performance. However, since the arguments were accessible and some apps may be unknowingly depending on private API, this change will follow the formal deprecation process. Alternative approaches for all three hooks are below:
```
locks: Since this documentation will be post-deprecation, I think we should slightly re-word this. Something in the spirit of http://emberjs.com/deprecations/v2.x/#toc_initializer-arity, maybe.

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -517,3 +517,95 @@ let author = this.store.peekRecord('user', 456);
...
+```
+##### `didReceiveAttrs` arguments
+
+Let's say you want to animate a thermometer widget showing the change between today's high and low temperatures.
+
+Before:
+
+``` javascript
+Ember.Component.extend({
+  didReceiveAttrs({ highTemp, lowTemp }) {
```
locks: You can't change this, or the destructuring won't work. `didReceiveAttrs` receives an object with `oldAttrs` and `newAttrs` attributes, so:

```javascript
didReceiveAttrs({ oldAttrs, newAttrs }) {
}
```

Is equivalent to:

```javascript
didReceiveAttrs(attrObj) {
  let oldAttrs = attrObj.oldAttrs;
  let newAttrs = attrObj.newAttrs;
}
```

Did you mean to destructure oldAttrs/newAttrs themselves, like `didReceiveAttrs({ newAttrs: { lowTemp, highTemp }}) {`?

source/deprecations/ember-data/v2.x.html.md
```diff
@@ -517,3 +517,95 @@ let author = this.store.peekRecord('user', 456);
...
+``` javascript
+Ember.Component.extend({
+  didUpdateAttrs({ highTemp, lowTemp }) {
+    if (highTemp !== lowTemp) {
+      this.thermometer.move({ from: highTemp, to: lowTemp });
+    }
+  }
+});
+```
+After:
```
locks: Can you add an empty line above "After:"?

locks:
@jenweber let me know if you need help resolving the deprecation

jenweber:
Thanks @locks. Just got back from the holidays and will update the PR tonight.

jenweber:
@locks let me know if anything else should be changed, or if I should resolve the conflicts too. Thanks!

jenweber:
@locks I think it's in the right place now. But I'm a bit confused by the "files changed" and "commits" associated with the PR (particularly some deletions that I didn't make and don't show up elsewhere), so please look them over and let me know if that looks normal. I haven't really worked on forks of actively developed projects.

PR #2772, opened 2016-12-21
adds g5 to ember-users

levibrown:
Adds G5 to the user list.


data/users.yml
```diff
@@ -886,6 +886,11 @@
   image: clark_x2.png
   alt: "Clark - Smart Insurance"
   use: "Clark's consumer ui is powered by Ember.js - available on our website and integrated into our iOS & Android apps."
+- url: https://www.getg5.com
+  image: g5_2x.png
```
locks: Heya, can you do two things for me?
- [ ] Rename g5_2x.png to g5_x2.png
- [ ] Include the smaller image, name it g5.png, and then put that in here.

Gracias :]

locks:
Thank you!

PR #2774, opened 2016-12-23
deprecate Ember.K

givanse:
deprecate Ember.K according to https://github.com/emberjs/ember.js/issues/14746

[rendered view](https://github.com/givanse/website/blob/0053f6e7f5d4556937e0b20dd5ca491c306f90cb/source/deprecations/v2.x.html.md#deprecations-added-in-212)

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -858,3 +858,38 @@ Index Content
 
 For more informations of how to use `ember-elsewhere`, please visit the official
 documentation [here](https://github.com/ef4/ember-elsewhere#ember-elsewhere).
+
+### Deprecations Added in 2.12
+
+#### `Ember.K`
+
+##### until: 3.0.0
+##### id: ember-metal.ember-k
+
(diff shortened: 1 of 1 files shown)


locks:
We'd like the explanation to lean much more in the direction of `--empty`. We think that most people aren't using the chaining property of `Ember.K` and it's preferable to go with the simpler option.
@givanse could you tweak it a little bit so that it's more "oh, here's the command with the `--empty` option. P.S. if you're using chaining, there's `--return-this`"?

givanse:
PR updated with added emphasis on the `--empty` flag
https://github.com/givanse/website/blob/d2b6d8d435dd7d156dc7b0e0c83b3af45f984651/source/deprecations/v2.x.html.md#emberk

givanse:
Header added:
https://github.com/givanse/website/blob/0053f6e7f5d4556937e0b20dd5ca491c306f90cb/source/deprecations/v2.x.html.md#deprecations-added-in-212

I don't know what to do about the failed deploy/netlify

> You don't have access to this build log.

locks:
@givanse it's not your fault, we're trying it out and all PRs are failing for some reason!

PR #2786, opened 2017-01-18
Update v2.x.html.md

arjunkathuria:
fix getOwner polyfill ember version comment.

cc:- @locks


locks:
Thanks!

rwjblue:
Would you mind also updating the example to the new polyfill syntax? Importing the module from ember-getowner-polyfill is deprecated...

PR #2788, opened 2017-01-18
Update v2.x.html.md

arjunkathuria:
move to new syntax, Importing the module from ember-getowner-polyfill is deprecated.

cc:- @rwjblue

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -335,7 +335,7 @@ After it installs, use as follows:-
 // Ember < 2.3
 import Ember from 'ember';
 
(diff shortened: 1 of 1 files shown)


locks:
@arjunkathuria given that you don't actually need to change the code with the new polyfill, I think it would be less confusing if we removed the last code block, and rewrite the paragraph a little bit to say you only need to install the addon. Thanks for the follow-up!

rwjblue:
This is looking good to me, but @locks should do a final review...

rwjblue:
@locks - Ping?

locks:
Thanks!

PR #2792, opened 2017-01-25
[WIP] New navbar

MelSumner:
Adding "learn" into the menu
Will transform into dropdown menus once the website can support RWD for the whole site.


source/stylesheets/_header.scss
```diff
@@ -1,188 +1,213 @@
-.header-nav.container {
-  width: 960px;
-  padding: 0 1rem;
-  list-style-type: none;
-  margin: auto;
+* {
```
locks: I already see this in site.css.scss, so should be unnecessary.

source/stylesheets/_header.scss
```diff
@@ -1,188 +1,213 @@
...
-
-.header-search {
-  margin-left: 4%;
-
-  .algolia-autocomplete {
-    display: block !important; // beats `inline-block` added by DocSearch script
-  }
-  .search-input {
+  padding: 0;
+  nav {
```
locks: Styling the `nav` tag directly seems undesirable, can we semantically have only one `nav` element in our page?

source/_header.html.erb
```diff
@@ -1,15 +1,43 @@
-<header class="header">
-  <nav role="navigation" aria-label="main">
-    <ul class="header-nav container">
-      <li class="header-logo">
-        <a href="http://emberjs.com/"><span class="visually-hidden">Ember homepage</span></a>
-      <%= link_to_page "Guides", "http://guides.emberjs.com" %>
-      <%= link_to_page "API", "/api" %>
+<script type="text/javascript">
```
locks: might want to wrap this in a `<% content_for :foot do %>`, see https://github.com/emberjs/website/blob/master/source/about.html.erb#L5-L23

source/stylesheets/_header.scss
```diff
@@ -195,4 +220,4 @@
   padding: 0 !important;
   position: absolute !important;
   width: 1px !important;
-}
+}
```
locks: Missing trailing whitespace ;P

source/stylesheets/_header.scss
MelSumner: - since it's scoped to the header it would be fine (hopefully we're not putting header elements with navs in them all over the place, but maybe that's something we should discuss)
- a site can have more than one nav; although if you do, you need to have an ID that helps differentiate them for accessibility purposes.

locks: Hm, I was thinking the footer links, and in-page sections like the Guides' table of content.

locks:
💥  let's get this party started

PR #2793, opened 2017-01-26
[wip] revert pull/2752 internal-api-docs

GavinJoyce:
reverts https://github.com/emberjs/website/pull/2752

This revert fixes (I think, I'm finding it hard to verify locally) the missing content in our API docs which I believe is our most pressing issue. The second most pressing issue is the missing API docs packages, which should be fixed with https://github.com/emberjs/ember.js/pull/14878.

Once these both land, we can regroup and implement `/internal` in a way that doesn't break the main API docs


locks:
@GavinJoyce thanks Gavin and sorry for dropping the ball so hard on this one. We should switch to `ember-api-docs` soon, so I'm closing this PR.

Issue #2716, opened 2016-11-04
A search with no results should indicate that there are no results


aren55555:
I just used the search bar on emberjs.com; I pasted in the query `modulePrefix `... nothing happened. I then press `Return`; and the page redirected to `http://emberjs.com/?`. I repeated this a couple of times thinking it was broken. 

Turns out there are actually no results for the `modulePrefix` query. However the search in no way indicates this.

acorncom:
Confirmed, I'm seeing this as well.  However, this seems to be an issue with the docsearch library (https://github.com/algolia/docsearch) that we're using. I'm seeing similar issues when searching on the React website for something that doesn't exist (https://facebook.github.io/react/).

@aren55555 Would you be interested in researching this further to see if you can work out a fix?  Would love to have it resolved, but time is pretty tight on this end.  If we don't find a volunteer, I'll probably just close this for now ...

aren55555:
@acorncom I'm not too familiar with docsearch; but it looks like they have closed a similar issue: https://github.com/algolia/docsearch/issues/92

Tonours:
Hi @acorncom a little fix for this issue is here #2799

aren55555:
@Tonours nicely done - much appreciated.

PR #2799, opened 2017-02-01
Update algolia version

Tonours:
Issue #2716 

- Updated algolia by modifiying CDN namespace version in order
to display an error message when no results found.

- Updated _docsearch.scss vendors accordingly to algolia CDN update
due to some markup update from Algolia.

Now when there is no result we have a UI feedback

screenshot

### DIFF
--- a/source/javascripts/app/docsearch.js.erb
+++ b/source/javascripts/app/docsearch.js.erb
@@ -2,7 +2,7 @@
     var script = document.createElement('script');
     script.type = 'text/javascript';
     script.async = true;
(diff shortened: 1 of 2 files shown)


acorncom:
@Tonours Much appreciated!  Would you have the time to make these changes to the https://github.com/emberjs/guides repo as well?  We're using the same setup in both places and I'd love to get those reflected over there ... 😀

Tonours:
@acorncom Yes, of course :-)

Issue #2796, opened 2017-02-01
Add deprecation page entry for `factoryFor` API.


rwjblue:
- [ ] Add to Ember 2.12 deprecations
- [ ] Update deprecation URL's in emberjs/ember.js to point to the final URL

---

The deprecation can suggest the usage of `ember-factory-for-polyfill` (if an addon only needs Ember 2.3+ support) or `ember-getowner-polyfill` (supports 1.10+) for addons.

For apps, things are somewhat simpler `getOwner(this)._lookupFactory('some:thing')` -> `getOwner(this).factoryFor('some:thing')`.

jenweber:
On it.

[RFC](https://github.com/emberjs/rfcs/blob/master/text/0150-factory-for.md) 

Notes to self: _lookupFactory was a private API mechanism, used primarily for passing in arguments when creating an instance of a factory. It's been replaced by the public API method factoryFor.

jenweber:
@rwjblue or @locks Do you prefer opening issues in each repo, or would you just reference this issue in commits on both repos? Just trying to generally gauge how you like to manage issues when the task spans multiple repos.

[PR submitted](https://github.com/emberjs/website/pull/2800)

rwjblue:
I'd just reference this meta issues from the Ember PR, but honestly it isn't something we are sticklers about anyways. We are mostly just happy for the help.

PR #2800, opened 2017-02-03
add deprecation guide for migrating from _lookupFactory to factoryFor

jenweber:
Please double check that this deprecation is under the right heading (added in 2.12).

Comments welcome!

[RFC](https://github.com/emberjs/rfcs/blob/master/text/0150-factory-for.md#example-deprecation-guide-migrating-from-_lookupfactory-to-factoryfor)
[Issue](https://github.com/emberjs/website/issues/2796)

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -314,7 +314,7 @@ do an ember install of the polyfill.
 ember install ember-getowner-polyfill
 ```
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you! Would you mind updating the deprecations in Ember to use the new URL (I believe they use TODO at the moment)?

jenweber:
Sure, I can do that

PR #2801, opened 2017-02-03
Fixed wording for the 2016 community survey

tzellman:
- less confusing to use percentages, since they are used elsewhere and reader doesn't have to do math
- fixed grammar


mixonic:
:clinking_glasses: thanks sir!

PR #2807, opened 2017-02-13
Motto img to text

MelSumner:
For @locks


locks:
Commit cherry-picked manually, danke 🗡

PR #2813, opened 2017-02-18
Added missing references to DS in model examples.

mthouve:
This adds missing DS references to DS.Model implementation examples.

Locations affected:

1. http://emberjs.com/api/data/classes/DS.Errors.html
2. http://emberjs.com/api/data/classes/DS.Model.html#property_attributes
3. 

http://emberjs.com/api/data/classes/DS.Model.html#property_transformedAttributes

### DIFF
--- a/data/data_api.yml
+++ b/data/data_api.yml
@@ -549,8 +549,8 @@ classes:
       import DS from 'ember-data';
 
       export default DS.Model.extend({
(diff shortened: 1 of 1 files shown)


locks:
Hi @mthouve, this documentation is generated from the Ember Data source code, so you'd need to change it there instead. Thanks for the effort!

PR #2830, opened 2017-03-10
Update urlForFindAll doc

ksin:
Noticed a typo while browsing some docs


locks:
Thanks for the contributing @ksin, but this is generated from the source, so it would need to be updated there. However, seems to be [already updated](https://github.com/emberjs/data/blob/c7151ecaed32f762756e9a1ee15b117837faeace/addon/-private/adapters/build-url-mixin.js#L146) so it's a matter of waiting for a release!

PR #2831, opened 2017-03-11
Fix typo

catto:
I found a typo when I read through the document, so I have fixed it.

### DIFF
--- a/data/api.yml
+++ b/data/api.yml
@@ -16651,7 +16651,7 @@ classitems:
     `_super`:
 
     ```app/routes/photos.js
-    import Ember from 'ebmer';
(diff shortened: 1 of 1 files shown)


acorncom:
@catto much appreciated! The files you updated get refreshed every release from the source files in the main Ember repo, would you mind making your change there too so we don't lose this change in the next release? Should be able to do that here: https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/route.js

Thanks!

catto:
@acorncom Oh, I see. I didn't notice that it is auto-generated. I am going to make changes there too!

PR #2836, opened 2017-03-16
Adds Rule of Three to ember-users

willviles:


locks:
@willviles hi! what happened here?

willviles:
@locks there was a merge commit, which I accidentally committed incorrectly... And now I realise force push undoing that was a bad idea.

locks:
It's merged now :D Thank you!

willviles:
@locks thanks!

PR #2847, opened 2017-03-22
Remove survey reminder + remove survey link from 2017 survey landing page

tzellman:
The 2017 survey runs until midnight (PST?) on 3/21 (tonight).

This PR removes the survey reminder that shows up on all pages, as well as the link (and related content) from the actual survey landing page.

Screenshot


rwjblue:
Just to confirm, we should land this PR after the survey has closed (sometime after 12:00am PST), right?

tzellman:
Yes, that is correct. We will make the survey private at some point in the early AM, so no worries if the link on the survey page stays active for a while longer. We likely won't be crunching numbers as soon as the clock strikes midnight. :)

PR #2875, opened 2017-04-03
Removed the 2017 survey link from the layout + related resources

tzellman:
This should be merged whenever it is decided it is a good time to remove the survey link from the top of the layout. This essentially reverts the [commit that added it](https://github.com/emberjs/website/commit/b2608c5bbfc625aba69b7246c872cabf68fa9fc0).

screenshot


locks:
💟

PR #2877, opened 2017-04-03
Extract common styles for Showcase link backgrounds

SaladFork:
This moves the styles used to create the rounded shadowed white pill backgrounds from being specific to the Ember Users page to be used for all showcase list links (currently just `/ember-users/` and `/sponsors/` as far as I can tell).

This fixes [the issue reported by @wifelette](https://github.com/emberjs/website/pull/2876#issuecomment-291246130).


rwjblue:
thank you!

Issue #2829, opened 2017-03-10
Left nav on the blog doesn't link to tags


mixonic:
The left nav, for example "Recent Posts" no longer links to the tags. Instead it collapses the menu (which seems less useful).

![screen shot 2017-03-10 at 11 11 25 am](https://cloud.githubusercontent.com/assets/8752/23809403/69d69ee2-0582-11e7-8302-c46e1829056c.png)

acorncom:
On individual blog pages (such as http://emberjs.com/blog/2017/03/19/ember-2-12-released.html), clicking the left sidebar headers will collapse and expand a given subsection as reported above.  However, when viewing either a main tag section such as Releases (http://emberjs.com/blog/tags/releases.html) or Recent Posts (http://emberjs.com/blog/tags/recent-posts.html) or the main blog page (http://emberjs.com/blog/) the sidebar headers don't collapse but instead link to the desired page.  We'd like to see all pages act like the main blog page does

PR #2883, opened 2017-04-05
slide-widget-head required for sliding TOC

mixonic:
should fix https://github.com/emberjs/website/issues/2829

Now we require the `slide-widget-head` on any links that are sliding widgets. The blog sidebar then does not use this class.


locks:
Thank you 🙏

PR #2884, opened 2017-04-10
[DEPRECATIONS] Add deprecation for ember-metal.map

Serabe:
Add deprecation for ember-metal.map, being deprecated in emberjs/ember.js#15119


locks:
@Serabe can you resolve the conflict, plox?

PR #2887, opened 2017-04-16
Add in the code for replacing /tomster with /mascot

yamboy1:
Might work, might not. Check before it goes up


locks:
Hi, thanks for the contribution but it does indeed not work. We need to clean up the routes beforehand, because there's /tomster/faq among others.

Issue #2862, opened 2017-03-28
example for didReceiveAttrs deprecation


asizer:
I found the "Before/After" example here not super illuminating:
https://emberjs.com/deprecations/v2.x/#toc_code-didreceiveattrs-code-and-code-didupdateattrs-code-arguments

The "After" code sort of looks like it's suggesting the old `highTemp` attr is available as `_highTemp` (or that there's no way to access old attrs, and you'll just have to pass everything in to the component in the future). To match the "Before" block, shouldn't it be something more like this?

```
didReceiveAttrs() {
  let oldTemp = this.get('_oldTemp');
  let newTemp = this.get('temp');

  if (oldTemp && oldTemp !== newTemp) {
    this.thermometer.move({ from: oldTemp, to: newTemp });
  }
  this.set('_oldTemp', newTemp);
}
```

The example here is more helpful, as it shows setting the incoming attr to an `_old...` attr for future comparison:
https://github.com/emberjs/rfcs/blob/master/text/0191-deprecate-component-lifecycle-hook-args.md#didreceiveattrs

locks:
Can you send a PR for this?

locks:
Thanks again for addressing this in #2892 :)

PR #2892, opened 2017-04-27
Changed example for "Arguments in Component Lifecycle Hooks" deprecation

asizer:
In the "Arguments in Component Lifecycle Hooks" section, changed the After example for `didReceiveAttrs` to better match the Before example, and more importantly, to show the manual saving of any properties one might want to compare old values and new values of during `didReceiveAttrs`. More details and link to inspiration for change in #2862


locks:
Appreciated @asizer!

PR #2852, opened 2017-03-27
Update ember source to 2.9

kiwiupover:
Remove deprecated Handlebars SafeString in favor of String htmlSafe

There are issues moving to glimmer2. 

In talk with @acorncom there are plans to move the home page of the ember website to an ember-cli fastboot app in the near future.

PR #2906, opened 2017-05-28
Clean up examples in about page

locks:
We're working towards a cleaner look.
Instead of having (outdated) code samples with little context,
we expect interested parties to check out the getting started guide.
We should also make sure the getting started is smooth sailing.

### DIFF
--- a/source/about.html.erb
+++ b/source/about.html.erb
@@ -66,54 +66,6 @@ responsive: true
   </div>
 </div>
 
-<div class="getting-started section">
-  <script type="text/x-handlebars" id="application">
-    <div class="example-tabs">
-      {{#each model as |item|}}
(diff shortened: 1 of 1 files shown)


wycats:
:+1:

stefanpenner:
whats holding up the merge?

locks:
@stefanpenner was waiting for someone else on the learning team to 👀, but we can merge.

acorncom:
@locks I'm ambivalent about the change, but don't mind doing so.  If we're going to do this, seems like we should remove the Ember app imported here https://github.com/emberjs/website/tree/master/source/javascripts/app/about and the `ember-middleman` Gem usage (seen in the update here https://github.com/emberjs/website/pull/2852)?

PR #2907, opened 2017-05-30
Mascot updates

wifelette:
Updating the FAQ, commission pricing sheet and request form to reflect addition of corporate mascot option. 

Also deleting/cleaning up an old page from this effort that never went anywhere.


source/tomster/commission.html.erb
```diff
@@ -65,26 +65,38 @@ title: "Tomster Commission Request"
               <span class="choice">
                 <input id="Field8_0" name="Field8" type="radio" class="field radio" value="Classic Tomster with Simple Shirt Change, $80" tabindex="5" checked="checked" required>
                 <label for="Field8_0">
-                  Classic Tomster or Zoey with Simple Shirt Change, $80
+                  UG Classic Tomster or Zoey with Simple Shirt Change, $80
```
acorncom: UG standing for user group I'm assuming?  Think it'd be worth spelling that out so it's clearer on this form?

wifelette: Ack, just an oversight, good catch!

wifelette: I was just trying to keep the line breaks tidy :p

acorncom: That's what I figured 😀 I vote for non-tidy 😛

locks: It doesn't look good ;X Can we explain beforehand and use UG and CO?

wifelette: @locks That works for me!

PR #2916, opened 2017-06-12
Fix borders on users

robustdj:
On the users page (`/ember-users`), the users in the leftmost column are missing their left border and the users in the rightmost column are missing their right border. You might have to resize the window width to see it.

screen shot 2017-06-12 at 10 47 27 am

This fix just adds a 1px padding to the container so that the borders always show.

screen shot 2017-06-12 at 10 50 40 am

### DIFF
--- a/source/stylesheets/site.css.scss
+++ b/source/stylesheets/site.css.scss
@@ -1607,6 +1607,7 @@ ul.showcase {
 
 #content ul.showcase {
   margin: 0;
+  padding: 0 1px;
 }
 
 #content .showcase li {


wifelette:
@robustdj Thanks for this!

@locks How do you feel about building some padding in, so that logos like noppo up above can't go that close to the edge? We have standard image sizes so it would shrink many of the current logos, but only by a smidge. Thoughts? 

Asking now because maybe we can talk @robustdj into playing around with it while in the code area ;)

locks:
@wifelette I'm not super into "punishing" the submissions that properly padded their images, but what do you say we merge this one and follow up with an issues/PR?

wifelette:
I'm happy to do the work of trimming the various existing pngs as needed, just seems silly to *keep* needing people to modify images for something so easily fixed in code :p It's been on my radar for a while. A new issue for this sounds great!

locks:
@robustdj thank you for your contribution! Would you like to send over another PR with the changes to the padding mentioned in the discussion above?
I've created an issue at https://github.com/emberjs/website/issues/2918.

PR #2921, opened 2017-06-19
Update Requirements section about Ruby version

tricknotes:
I removed the reference to Ruby 1.9 and 2.0 because they are too old to run website :<


locks:
Thank you!

PR #2927, opened 2017-06-22
Extract footer to glimmer-styleguide

locks:
Although they are components to be used in Emberjs family of web resources, the package is prefixed "glimmer" because the components are implemented in Glimmerjs.
We are using the web component shim for easy integration.

### DIFF
--- a/source/layout.erb
+++ b/source/layout.erb
@@ -23,6 +23,7 @@
     <% if current_page.data.no_index.eql? true %>
       <meta name="robots" content="noindex">
     <% end %>
+    <link rel="stylesheet" href="https://glimmer-styleguide.global.ssl.fastly.net/glimmer-styleguide/master/app.css" />
     <%= yield_content :head %>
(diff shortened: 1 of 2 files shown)


source/layout.erb
```diff
@@ -68,5 +69,6 @@
         var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
       })();
     </script>
+    <script src="https://glimmer-styleguide.global.ssl.fastly.net/glimmer-styleguide/master/app.js"></script>
```
acorncom: Should we use a simple helper for all these urls so it's easier to adjust when we inevitably need to in the future?

acorncom:
@locks is this ready to merge?

locks:
Yup!

source/layout.erb
locks: sounds like a good idea

PR #2928, opened 2017-06-22
Drop unused gems

tricknotes:
They are unnecessary since https://github.com/emberjs/website/commit/2b192878e77816d8274a2179e00aa4c9e7ca4c98


locks:
👍 looking good.

PR #2929, opened 2017-06-22
Compress company logos

### LINKED ISSUE #2923: Huge amount of image data for ember users page
Just stumpled upon the fact, that the aggregated file size of all logo images on https://emberjs.com/ember-users/ is pretty huge, even on my desktop machine it took a while to load (>6s), probably quite unusable on a mobile phone with a slow connection:

image

A lot of the logos are way too large given their small size, some of them seem to include unnecessary meta data (Adobe Fireworks stuff, maybe color profiles as well). But even the smaller ones could be compressed much better without a noticeable loss of quality, by converting them from raw PNG24 to optimized PNG8.

This could be done manually for the existing files (e.g. by using some service like https://tinypng.com/) and requiring new submissions to do the same, or by adding some optimization step to the build process (running some CLI tool like https://pngquant.org/). (*I have no idea about Ruby and Middleman, so I cannot be of any help there...*)

simonihmig:
Processed all logos with https://imageoptim.com. Image quality setting was 80%, so optimization was not lossless, but should not be noticeable.

Fixes #2923

Overall files size was reduced from 5.2MB to 1.9MB!

@wifelette @locks The optimization run did take quite some time. So having the uncompressed images in the repo and doing the optimization at build/deploy-time is probably not a good idea in retrospect! 😬  I added a notice to the `users.yml`, so new additions should be submitted already optimized. Maybe you should check this for new PRs, at least by looking at the filesize (most logos are actually < 5KB now)

Fwiw I think there are a lot more cases of images that could benefit a lot from an optimization run, spread throughout the guides and the other pages! Especially PNG24 files are huge in general. 
The imageoptim is pretty easy to use, so I think we should optimize them as well! I can come up with another PR if that sounds good to you...


locks:
> Fwiw I think there are a lot more cases of images that could benefit a lot from an optimization run, spread throughout the guides and the other pages!

We've been talking about this. Ideally we'd want the images somewhere such that they could be referenced from wherever necessary, I'd say most don't need to be fingerprinted.
We should keep R&Ding.

wifelette:
@simonihmig —thanks so much for doing the work on this!

simonihmig:
@wifelette no problem! :)
@locks is this related? I was more referring to the fact that there were pngs around that would benefit form compression, even when not changing their file location. 

Other than having my machine's CPU be busy for some minutes, there is not much work involved in compressing all images. Should I open another PR for all the other images? (taht is compressing them without any relocation)

locks:
> Should I open another PR for all the other images?

👍

> @locks is this related? I was more referring to the fact that there were pngs around that would benefit form compression, even when not changing their file location.

Yeah, I was referring to what we should do going forward. If we used a asset host, we could likely treat new images as they come in, instead of running the (slow) compression on build of the website.

PR #2936, opened 2017-06-30
Update Ruby version to 2.4.1 with some gems update

tricknotes:
json@1.8.3 gem couldn't be built on Ruby 2.4.1 caused by incompatible with internal Ruby changes.
Also, other some gems show the following warning:
```
warning: constant ::Fixnum is deprecated
```

So I updated Ruby and some gems version.


locks:
Thank you for your work!

PR #2937, opened 2017-06-30
Remove header materials

tricknotes:
It seems these files are extracted as [glimmer-styleguide](https://github.com/ember-learn/glimmer-styleguide) by https://github.com/emberjs/website/pull/2930.


locks:
Thanks :)

PR #2940, opened 2017-07-01
add draft of ember-data blog post

runspired:


stefanpenner:
things i believe are part of this release, we may want to talk about:

* support for babel-preset-env targets when used \w new versions of ember-cli (basically using babel 6)
* fix verbose/slow babel output
* Batch record arrays -> https://github.com/emberjs/data/pull/4849
* move some machinery only used for assertions to devOnly
  * https://github.com/emberjs/data/pull/4948

PR #2941, opened 2017-07-02
Clean up CSS

locks:


source/stylesheets/site.css.scss
```diff
@@ -34,7 +34,6 @@ $small-max-width: 767px;
 }
 
 body {
-  background-color: #faf4f1;
```
MelSumner: a background-color on body should be defined, even if it is white.

locks: Added back in.

PR #2973, opened 2017-08-07
Updated Ember user logos

ciolt:
I noticed that some of the logos on the [**users page**](https://emberjs.com/ember-users/) were outdated. So, I went out and updated the logos of these companies:

- TED
- Travis CI
- Zendesk
- Groupon
- Square

I also moved some of the more recognized companies to the top of the users list. Don't panic! I'll explain why. When I first found Ember, I noticed some of the featured users on the homepage are companies that I and a lot of people have never heard of.

> image
> It's not very convincing. Meanwhile, Microsoft, LinkedIn, and TED are hidden somewhere on another page...

GitHub for example features a lot of well-known companies. Including Google, Facebook, NASA, Walmart, and Spotify. They don't include small tiny companies like Jaylow Software—which nobody really has ever heard of.

> image
> It shows that GitHub is used and trusted by big brands and well-known companies.

I think this could definitely attract new users (and companies!) to using Ember. Companies really like it when they know they're buying into projects used by trusted organizations. It also becomes easier to convince your boss to use it.


ciolt:
Also, the company Urbanspoon no longer exists (they were acquired by Zomato)

Should their user status just be updated and renamed to Zomato, or should they just be removed altogether? I don't think keeping a dead company on the list is the brightest idea.

pixelhandler:
@ciolt just curious, I wonder if changes to those logos need some approval from the owners of the trademarks. Also it looks like the sizes are different in this PR see the diffs of the graphics, does that affect the display? I'm not sure about changing the order I think @wifelette would need to provide insight into the changes and ordering of these logos for companies using Ember. Thanks for your help I'm sure there are some needed updates just will need to figure out the details still.

wifelette:
Confirm, the logo changes would need approval. As for reorganizing the list, I review periodically, and probably should again, so this is a good reminder :) We balance between companies that are large and impressive, and companies that supported Ember early on.

All that said, I do like the idea of a monochrome grid like GH does, versus the carousel right now, as IIRC the carousel brought up accessibility issues we didn't have solid solutions for yet. @MelSumner, am I remembering that right?

If someone wanted to put together a design like that for us, I could commit to reaching out to the companies we wanted to feature to get new monochrome logos + approvals.

wifelette:
@ciolt Note Issue #3011 —finally back to thinking about this one, thanks again for the reminder!

That's just the tidying part though. I'd still love for someone to pitch a design for monochrome logos on just the home page.

wifelette:
Closing this for now. Some of it happened in #3048. The logo _updates_ would still be good, but I'll need to reach out to the original submitters. It's on the list!

PR #2978, opened 2017-08-19
Add Icicle to Ember Users

wicliff:
We are using Ember for nearly two years now. Built front-ends for multiple products using this technology.


wifelette:
Did _not_ verify that the code or logo are correct, but email received!

Has this been staged? The logo didn't look to be the right size in a quick glance, but my eyes might be playing tricks on me.

wicliff:
Great!

The logo is 420x206 as per the instructions. Let me know if there is a need for a different size for this.

locks:
Thank you for using Ember 😁

PR #2985, opened 2017-09-01
Correct link to PhantomJS maintenance announcement

backspace:
Hey, I noticed this link was the same as the previous one, I think it probably was meant to be a link to the announcement. 😀


locks:
Thank you!

PR #2995, opened 2017-09-26
WIP | RFC 236: Ember.String loc, fmt, prototype extensions, and module api

locks:


jenweber:
@serenaf looks like these should be rolled into your deprecations app at some point?

serenaf:
@jenweber Good point , will move them over to deprecation-app as well )

jessica-jordan:
@locks @serenaf @jenweber Is this safe to be closed after moving these over to the deprecations app?

PR #3012, opened 2017-10-10
Put Helvetica Neue back in for better iOS

MelSumner:
Since iOS doesn't render Google Fonts very well at all, putting Helvetica Neue first should help resolve some issues where Source Sans Pro looks great on any other device but iOS.


MelSumner:
@locks

Issue #3018, opened 2017-10-10
og:image and twitter:image point to 404'd images


chadhietala:
A lot of social media sites will look at these to take to display an image for the page. This seems to be broken.

<img width="1354" alt="screen shot 2017-10-10 at 7 33 38 pm" src="https://user-images.githubusercontent.com/183799/31415535-0eebe0f0-adf2-11e7-9f59-56086351fb27.png">
<img width="1611" alt="screen shot 2017-10-10 at 7 33 49 pm" src="https://user-images.githubusercontent.com/183799/31415536-1158220e-adf2-11e7-9966-9059ea348b64.png">

This ends up looking like this:

<img width="553" alt="screen shot 2017-10-10 at 7 33 58 pm" src="https://user-images.githubusercontent.com/183799/31415543-1b05b50a-adf2-11e7-97cf-599b1be4e162.png">

NullVoxPopuli:
what is the image supposed to be?  just a small screenshot of the article?

chadhietala:
@NullVoxPopuli I think its supposed to be Tomster.

NullVoxPopuli:
@chadhietala yeah, I think middleman has some limitations around the use of images in meta tags, so I just excluded the image(s) from the asset fingerprinting

acorncom:
Closed by #3020

PR #3020, opened 2017-10-11
remove the two tomster images from the asset_hash for social use.

NullVoxPopuli:
addresses: https://github.com/emberjs/website/issues/3018

looks like middleman just doesn't want to deal with urls in meta tags.
though based off this: https://docs.datocms.com/middleman/seo.html there may be a gem or something to more properly handle fingerprinted full-url assets ref'd in meta.


acorncom:
Thanks!

PR #3025, opened 2017-10-11
change api docs link in blog post to one that works

toddjordan:
With the current link situation https://emberjs.com/api works while https://emberjs.com/api/ember does not.

Lets change the link in the blog post so people can see the new docs...


toddjordan:
cc @locks @mixonic

Issue #3029, opened 2017-10-12
Remove duplicate rules from stylesheets


locks:
Due to some duplication in the SCSS files, we have several rules that are duplicated, for example https://github.com/emberjs/website/blob/master/source/stylesheets/base/_variables.scss#L37 and https://github.com/emberjs/website/blob/master/source/stylesheets/base/_variables.scss#L53.

The task is to clean these up, also removing any commented out styles.

zzbazinga:
I would like to take this one on :)

PR #3031, opened 2017-10-12
Remove duplicate rules from stylesheets #3029

zzbazinga:
Removed duplicated rules and delete unnecessary comments.


locks:
Thanks for contributing @zzbazinga, keep them coming!

PR #3014, opened 2017-10-10
Update font-family value

MelSumner:
Updated font-family on body element to use $base-font-family variable for better consistency.

PR #3037, opened 2017-10-14
Fix wrong font-family on blog

oskarrough:
Removes an extra font-family on the blog. With this change, it'll use Source Sans Pro, like the rest of emberjs.com.

### DIFF
--- a/source/stylesheets/blog.css.scss
+++ b/source/stylesheets/blog.css.scss
@@ -121,7 +121,6 @@ body.blog.responsive {
 }
 
 article.blog-post {
-  font-family: "Helvetica Neue", "Helvetica", "Roboto", "Arial", sans-serif;  
   h1 {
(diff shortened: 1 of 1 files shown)


locks:
cc @MelSumner

MelSumner:
@oskarrough @locks 
We can't do this one quite yet- we need to set the base-font variable correctly first- 
```
$base-font-family: "Helvetica Neue", "Source Sans Pro", Helvetica, sans-serif;
```
And then, in the body element we need to set ```font-weight: 400;```
Then we can remove this line of CSS. 

Reason why: iOS does not handle Google Fonts very well at all.

locks:
https://github.com/emberjs/website/pull/3014 is merged.

oskarrough:
So we can merge now? @MelSumner

PR #3038, opened 2017-10-15
Dockerize

NullVoxPopuli:
Just copys from the guides repo.

Since all the setup stuff is going to be the exact same, instead of  copying all the readme pieces everywhere, should we have a central location where all the middleman/website docs are?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM dockerizedember/website-dev:latest
+FROM emberjs/website-dev:latest
 
 # Customizations for this project here.
 # e.g.: if you have dependencies in addition to what the base image provides:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM dockerizedember/website-dev:latest
+
+# Customizations for this project here.
+# e.g.: if you have dependencies in addition to what the base image provides:
+ADD Gemfile /src/Gemfile
+ADD Gemfile.lock /src/Gemfile.lock
+RUN bundle install
+
+ADD . /src
--- a/README.md
+++ b/README.md
@@ -7,6 +7,10 @@ The website for the Ember.js project.
 
 To get started:
 
+## Running locally with Docker (recommended)
+
+## Running locally with Ruby and Middleman
+
 ``` sh
 git clone https://github.com/emberjs/website.git
 cd website
--- a/README.md
+++ b/README.md
@@ -9,6 +9,28 @@ To get started:
 
 ## Running locally with Docker (recommended)
 
+This is the recommended method for new contributors.
+Although the website is built with Ruby, most work is done in Markdown files.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -7,6 +7,10 @@ The website for the Ember.js project.
 
 To get started:
 
+## Running locally with Docker (recommended)
```
locks: Aren't you missing the instructions?

NullVoxPopuli: yup. lol

NullVoxPopuli: @locks added

README.md
```diff
@@ -7,6 +7,32 @@ The website for the Ember.js project.
...
+
+This is the recommended method for new contributors.
+Although the website is built with Ruby, most work is done in Markdown files.
+You don't need to know Ruby or install its dependencies to help out. Simply follow
+the Docker container instructions below to install and run locally.
```
locks: no "simply" ;P

NullVoxPopuli: :+1:

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM emberjs/website-dev:latest
+
+# Customizations for this project here.
+# e.g.: if you have dependencies in addition to what the base image provides:
+ADD Gemfile /src/Gemfile
+ADD Gemfile.lock /src/Gemfile.lock
```
springerigor: In general `COPY` is a recommended way to copy files into a container.

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN bundle install
```
springerigor: A nice tip is to use `bundle check` before calling `bundle install`-> `bundle check || bundle install`

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM emberjs/website-dev:latest
+
+# Customizations for this project here.
+# e.g.: if you have dependencies in addition to what the base image provides:
```
springerigor: I am not convinced if the above comment provides much value, maybe it would be better to state here why `Gemfile*` are copied into a container?

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: '2'
```
springerigor: Maybe `3.4` as it is the most recent one?

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+services:
+  guides:
+    build:
+      context: .
+      dockerfile: ./Dockerfile
```
springerigor: `Dockerfile` is a default one so this line can be skipped.

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+    command: bundle exec middleman
+    ports:
+      - "4567:4567"
```
springerigor: `- "4567"` is enough

jenweber:
Dismissing locks' review as the requested changes have been addressed.

PR #3063, opened 2017-10-29
updated call to action on home page

MelSumner:
**Changes Made:**
- set the CTA so it doesn't take up the full-width on mobile. 
- set the CTA box at an increased width on desktop (350px -> 425px)
- made the heading text a darker gray for better contrast
- increased the font-size, padding and overall width for the CTA button
- alpha-sorted CSS properties
- increased the font-size & adjusted the line-height for the CTA text
- updated the CSS so the bottom of the CTA shows on mobile (it was being hidden)

**Screenshots:** 
Desktop view: 
pasted image at 2017_10_29 04_24 pm

Mobile view: 
pasted image at 2017_10_29 04_32 pm

### DIFF
--- a/source/stylesheets/components/call-to-action.css.scss
+++ b/source/stylesheets/components/call-to-action.css.scss
@@ -3,24 +3,27 @@
 .call-to-action {
   background-color: white;
   margin: 3em auto;
+  max-width: 90%;
   text-align: center;
 
   a {
-    color: #D0351A;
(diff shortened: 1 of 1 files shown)


MelSumner:
@wifelette that may be due to the quality of the screenshot- but we can iterate if it does look like that once the changes are live. Definitely still want it to look nice!

PR #3080, opened 2017-11-27
small style tweaks based on team feedback

MelSumner:
Made some small tweaks based on team feedback: 
- removed uppercase for h2
- removed bottom padding for h1 (the margin is fine)
- removed background color from user logos (left the border and the hover shadow)
- removed text-transform:capitalize from the h1
- added a ".no-capitalize" class to the ember-users h1 just in case someone changes the CSS again before we have the chance to clean it up.

### DIFF
--- a/source/ember-users.html.erb
+++ b/source/ember-users.html.erb
@@ -3,12 +3,12 @@ title: "Users"
 responsive: true
 ---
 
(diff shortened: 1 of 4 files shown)


jenweber:
Thanks Mel!

PR #3084, opened 2017-12-02
CSS Cleanup

MelSumner:
- tried to eliminate some of the extra top margins/padding on pages
- replaced more colors with appropriate variables (i.e., white was replaced with $white) 
- removed $orange-color (replaced instances with $orange or $orange-darker, where appropriate) 
- removed $ember-orange (replaced with $orange) 
- updated the value for white (fff  to fdfdfd)
- updated the value for black (000 to 212121)
---------

### DIFF
--- a/source/stylesheets/about.css.scss
+++ b/source/stylesheets/about.css.scss
@@ -17,7 +17,7 @@
 
     a {
       background-image: none;
-      color: white;
+      color: $white;
       letter-spacing: 1px;
       min-width: 100%;
       display: block;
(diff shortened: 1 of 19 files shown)


MelSumner:
It's a little tricky since I'm on Windows, but having the Heroku logs helps me determine if the build was successful or not. This is why you'll see extra commits during this CSS cleanup work. Once we're not using middleman anymore, this issue will be resolved (since I'll be able to check that a build was successful locally).

PR #3088, opened 2017-12-04
updating CSS variables to prep for CSS cleanup

MelSumner:
- changed background color to $white
- added status colors
- added standardized gray shades
- grouped colors for deprecation (including browns)

Screenshot of colors: 
image


MelSumner:
Since this is not removing any code and only adding new variables for use, I'm merging it.

PR #3095, opened 2017-12-08
Add missing href for Ember Conf CTA

HeroicEric:
I'm not sure if this is the URL that this is supposed to go but I think this is better than not going anywhere 😄 

LMK if this should point elsewhere.


acorncom:
Agreed, that certainly looks like the link is needed. 😀 Thanks @HeroicEric.

@jeffdaley I'm assuming this was just a mistake, not something we were doing deliberately?

jeffdaley:
Good catch! That button is supposed to link to emberconf.com

I've updated that (along with a CSS tweak) here https://github.com/emberjs/website/pull/3096

PR #3099, opened 2017-12-11
Add upcoming deprecations

locks:
Complements https://github.com/emberjs/ember.js/pull/14590.

### DIFF
--- a/source/deprecations/v2.x.html.md
+++ b/source/deprecations/v2.x.html.md
@@ -1090,15 +1090,22 @@ export default Controller {
 }
 ```
 
-#### Ember 2 Legacy
+### Deprecations Added in 2.18
+
+#### `targetObject`
+
+##### until: 3.5.0
+##### id: ember-runtime.using-targetObject
+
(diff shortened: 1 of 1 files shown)


serenaf:
Original PR for reference :) https://github.com/emberjs/ember.js/pull/14590

serenaf:
@locks According to this https://www.emberjs.com/blog/2018/01/01/ember-2-18-released.html#toc_changes-in-ember-js-2-18 it is since 2.18.0 and until 3.5.0

jenweber:
@locks can you remove backticks around the id? It is a small thing but makes future CSS easier. Also while you're at it, can you promote `Ember 2 legacy` up a level, so three `###` instead of four? It ought to be its own section now that I look at it.

I totally used `targetObject` in my apps. Oops.

locks:
@jenweber done

jenweber:
I noticed an issue with using backticks in the deprecation title too, but it's a styling problem that affects other listings so I'll open a separate PR/Issue

PR #3113, opened 2017-12-27
Revised EmberConf CTA

jeffdaley:
- Added "register" language
- Made entire banner clickable
- Misc. restyling


MelSumner:
Is this still an active PR? It looks like there are some conflicts to resolve.

PR #3118, opened 2018-01-03
programmatically adding deprecations

serenaf:
cc @locks 
This is a WIP Pr to programmatically generate the deprecations page. The general idea is to create an individual markdown file for each deprecation, with name, id, since and until in frontmatter.

There are 3 added helper methods in config:
- sub_pages(dir) retrieves all resources in a specific sub directory
- pages_by_version uses above method to retrieve a subset of resources and then groups them by id
- get_content displays markdown information (needed to display the content of each individual file)

Note this is a WIP, so not all deprecations have been translated to the new format, i just want to get a direction update.

Each version would have a .erb.html file that then uses the new helper methods to display the information 

Current problems/needs work:
- Ordering of the TOC (e.g items of 2.11 would come after 2.1 and not after 2.9)
- Generation of bookmark code


serenaf:
Closing this PR in favour of https://github.com/serenaf/deprecation-app
Middleman compiles all .mdf files in the src folder to pages, which is not desirable. We are making good progress with the above Ember app, which will make contributing much easier as well :)

PR #3123, opened 2018-01-05
Link to rfc mdbook

sivakumar-kailasam:
**NOTE:** To be merged after https://github.com/emberjs/rfcs/tree/rfc-process-update is merged to its master branch

### DIFF
--- a/static.json
+++ b/static.json
@@ -7,6 +7,9 @@
     },
     "/statusboard": {
       "origin": "https://ember-learn.github.io/statusboard/"
+    },
+    "/rfcs": {
+      "origin": "https://emberjs.github.io/rfcs/"
     }
   },
   "error_page": "404.html",


static.json
```diff
@@ -7,6 +7,9 @@
     },
     "/statusboard": {
       "origin": "https://ember-learn.github.io/statusboard/"
+    },
+    "/rfcs": {
+      "origin": "https://emberjs.github.io/rfcs/"
```
acorncom: This url seems to fail on my end, is that happening for you too?

locks: DNS for emberjs.github.io is stealing the URL.

sivakumar-kailasam:
@locks I'll close this for now since this problem isn't going away right now. But I'm thinking if we should adapt the pattern of displaying the rfcs inside the main ember app similar to how our blog posts work

PR #3124, opened 2018-01-06
updating gemfile to support windows dev environments

MelSumner:
- added platform support for windows
- downgraded eventmachine gem to specific version
- added :x64_mingw platform to gems tzinfo-data and wdm


MelSumner:
Looks like this will make this work for me on windows, but break things otherwise, so I'll probably have to keep this local (unless someone knows otherwise). 

> WARNING:
>        Removing `Gemfile.lock` because it was generated on Windows.
>        Bundler will do a full resolve so native gems are handled properly.
>        This may result in unexpected gem versions being used in your app.
>        In rare occasions Bundler may not be able to resolve your dependencies at all.
>        https://devcenter.heroku.com/articles/bundler-windows-gemfile

locks:
I rebased and realized that master already included these changes via the navbar!

Issue #3082, opened 2017-11-30
Update navbar


MelSumner:
**To Do:** 
Implement navbar into the current website . 

**Prototype:** 
https://codepen.io/melsumner/pen/b09957277ef32aa0626bf6c227a4e961 

Confirm that it works in all major browsers, and is responsive.

jkumar712:
I shall take it up! Should it be change in `ember-learn/glimmer-styleguide` or add a partial ?

MelSumner:
I think partial, because the current Glimmer one isn't rendering in IE.

jkumar712:
I am working on. Will raise PR within this week.

locks:
By partial do you mean middleman partial?

jkumar712:
@locks Yes, I asked in the mean middleman partial.

samselikoff:
Dropdowns look great on the new navbar!

I felt like the "ember" logo is a bit tight/needs more whitespace, wanted to come here and get yall's thoughts? 

Also what about the gradient (the one that's still on https://emberjs.com/api/ember/2.18/modules/@ember%2Fapplication), I felt like that was a nice touch

acorncom:
@MelSumner Thoughts about Sam's comments here? I'd be inclined to see the nav bar get a tad taller myself so it looks a bit more roomy in the header. I like the gradient as well, but @samselikoff that may have been removed for accessibility reasons in the interests of making things more distinct ...

MelSumner:
@samselikoff @acorncom we could iterate on it- the gradient on the header is something that went away based on feedback that was exactly the opposite. Additionally, there wasn't enough contrast to pass accessibility tests.  

If anyone wanted to help by submitting a PR or comment with a CodePen mockup ([you could just fork the styles in progress if you wanted](https://codepen.io/melsumner/pen/9d551738a81e319a773395a2cfa1a82e)) for https://github.com/ember-learn/ember-styleguide/issues/3 that would probably be the most helpful. The current iteration was only meant to be a stopgap while we transform the parts into components. 

Thanks!

PR #3128, opened 2018-01-07
more style cleanup & responsive conversion work

MelSumner:
tons of updates in this one, but there is/was something going on that I don't understand yet, so it's still a WIP. 

@locks if you were curious and wanted to pull down the branch and see how it builds, you maybe would like what you see (if it works).


locks:
@MelSumner review app seems to have deployed properly

MelSumner:
I think this is deployable, but could use some help to determine how (if at all?) this will affect the guides. 

After this is merged, I'll move on to #3138, then #3082.

PR #3142, opened 2018-01-08
[WIP] feature/#3082

MelSumner:
Still to do: 
- [ ] consolidate header/nav CSS (& eliminate the stuff that isn't used)
- [ ] add JS for toggles/dropdown
- [ ] sort out the search 
- [ ] add in the real URLs
- [ ] adjust the jobs tag (it'll need to be moved up)


locks:
what happened?

MelSumner:
I was trying to do too much in one PR.

PR #3150, opened 2018-01-11
patch for missing includes

MelSumner:
this patch addresses issues introduced when gems were bulk updated. It should be considered a temporary fix; long term solutions will be addressed with CSS upgrades.


MelSumner:
solved with #3153 3153

PR #3151, opened 2018-01-11
Add Khorus software to the list of ember users

chbonser:


wifelette:
@chbonser The image _looks_ like it might be wider than specified in the Yaml file. Can you verify?

chbonser:
@wifelette good catch. Image size has been adjusted.

locks:
@chbonser thanks for the update, however, making the image be 210px wide instead of 420px means it will look blurry in "retina" displays. Do you want to resubmit before merging so it looks crisp for everyone?

locks:
Thank you @chbonser :)

PR #3157, opened 2018-01-15
header/new navbar initial commit

MelSumner:
Initial commit that adds the re-vamped navbar to the site that is keyboard navigable. 

**Screenshot:** 
Desktop:
ember-navbar
Mobile: 
image

**Still to-do:** 
- Remove bourbon and neat from build (will be done in a [separate issue](https://github.com/emberjs/website/projects/2#card-6708469))
- [x] someone on a mac will have to replicate the Gemfile and Gemfile.lock since Heroku will kick it back because it was generated on a PC (also do we know why it rejects PC generated Gemfiles??) 
- once the sitemap is updated, the navbar code should be updated to use the Middleman helpers (https://github.com/emberjs/website/projects/2#card-6708506) 
- [ ] this navbar will need to be added to guides & api (@jenweber & @toddjordan)


jenweber:
@MelSumner just to be clear, what if any blockers are there before this can be merged? Just propagating the header to guides and api docs?

jenweber:
@MelSumner One thing I noticed is that at some point we have clobbered the search styling. This is not an issue introduced in this PR but some past work, since it is present in emberjs.com home page as well
screen shot 2018-01-20 at 2 50 50 pm

jenweber:
Tested in Safari, Chrome, and Firefox. 

When using the mobile/hamburger menu, the search results are not visible. To reproduce, try typing in the search box and with your cursor in the menu, scroll down:
screen shot 2018-01-20 at 3 14 24 pm


Very slight differences in font weight and dropdown menu color between Firefox (top) and Chrome/Safari (middle/bottom). I do not think this should block, maybe just a subsequent PR for the ember-styleguide if we even care. Not sure why the color is different when the hex value is the same.

screen shot 2018-01-20 at 3 08 35 pm

locks:
@jenweber likely due to the color profile being mis-calibrated.

MelSumner:
@jenweber I'm not sure how to get around the Heroku thing- it's rejecting my changes to the Gemfile and Gemfile.lock because they were generated on Windows. Maybe someone can PR this branch from a macbook and see if Heroku is okay with that?

MelSumner:
> Very slight differences in font weight and dropdown menu color between Firefox (top) and Chrome/Safari (middle/bottom). I do not think this should block, maybe just a subsequent PR for the ember-styleguide if we even care. Not sure why the color is different when the hex value is the same.

I don't think it has to look the same in all of the browsers anyway. Browsers are weird and sometimes do stuff that conflicts with other browsers (historically, some of them are on purpose but that's a different convo).

MelSumner:
> One thing I noticed is that at some point we have clobbered the search styling. This is not an issue introduced in this PR but some past work, since it is present in emberjs.com home page as well

okay- that's most likely a side-effect of trying to DRY up the CSS. thanks for pointing it out!

MelSumner:
> When using the mobile/hamburger menu, the search results are not visible.

definitely unintended.

MelSumner:
Updates per feedback:
1. Changed search dropdown highlight color (per Jen's feedback): 
image

2. Changed position of "Deprecations" in navbar (per Yehuda's feedback): 
image

MelSumner:
@jenweber fixed the mobile search issue. @locks once this finishes building, if all goes well it should be okay to merge. :+1:

MelSumner:
Update to visual design on search: 
image

search suggestions being displayed on a small viewport: 
image

jenweber:
Layout bug - the search results obscure the search bar
screen shot 2018-01-27 at 6 24 47 pm

MelSumner:
Okay, resolved it (and made sure it was still fine on smaller viewports): 
image

jenweber:
It appears that edge has a default gray placeholder that was handled in the old styling, but not handled here:
screen shot 2018-01-28 at 12 00 08 pm

MelSumner:
I had to take out the search icon and put it back on the left as a background image because of the conflicts of the base styles and Algolia. The CSS is a little specific because of IE and Edge re: the placeholder text color. 

After checking to ensure the previous browsers still worked, I used browserstack to test all of the changes in Edge, here's a screenshot: 
image

If there are no further change requests, I'll merge.

PR #3163, opened 2018-01-20
Update Heroku logo and add x2 version

crcastle:
Updating the Heroku logo to the current version as specified [here](https://brand.heroku.com).  Also adding an x2 version for high DPI screens.


jenweber:
Hi @crcastle, thanks for the update! Looks nice and crisp :) It looks to me like we maybe need to update the footer icon too - can you confirm?

screen shot 2018-01-20 at 5 49 32 pm

crcastle:
Hi @jenweber, the footer icon is good! I was mainly concerned about the lowercase "heroku" on the /ember-users page.  We've moved to only uppercase when the name appears along with the logo. Thanks!

PR #3171, opened 2018-01-24
Improved color contrast on Emberconf CTA button background

### LINKED ISSUE #3164: Insufficient Color Contrast: "Learn More" button
**Context:** 
The EmberConf CTA "learn more" button does not have sufficient color contrast to pass WCAG 2.0 AA standards. Users who are color-blind would find it difficult, if not impossible, to distinguish between the button background and the text. 

**To Fix:** 
Ensure that the solution has a color contrast ratio of 4.5:1. Color contrast tools (like https://webaim.org/resources/contrastchecker/) can help with this. 

**Screenshot:** 
image

kennethlarsen:
Resolves #3164 

How it looks with the new color:
screen shot 2018-01-24 at 19 33 20

### DIFF
--- a/source/stylesheets/components/ember-conf.css.scss
+++ b/source/stylesheets/components/ember-conf.css.scss
@@ -69,7 +69,7 @@
   }
 
   a {
-    background: #4AA0B8;
+    background: #00829a;
     border-radius: 0.25em;
     color: white;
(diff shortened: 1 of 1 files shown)


wifelette:
Thanks so much for tackling this!

Issue #3156, opened 2018-01-15
Update content: Community Page


MelSumner:
**Context:** 
The content should be occasionally reviewed for accuracy, and updated if not. 

**To-do:** 
Review the content for the community page on the website, and update the content if outdated or inaccurate. 

**Screenshot (of section that might need updating):**
![image](https://user-images.githubusercontent.com/4587451/34954741-32373488-f9e7-11e7-8085-34e08e510ade.png)

jenweber:
I had a UX designer look it over (aka my partner). Recommendations:
- Remove the video or put it at the bottom
- Simplify header text: "Stuck? Lost? Get Help..." should simply be "Discussion" with a list of the various venues (bullet points for simplest option, icons and descriptions to be fancy)
- Simplify newsletter heading. Should just say "Newsletters"
- Simplify "Something Fishy" to Report an Issue

jenweber:
Results of UX poll:

jenweber [Yesterday at 3:21 PM]
in #general
Does anyone have a second to test an open source website's UX? I'm wondering how hard it is for you to find a link to the discussion boards or slack. Where did you look before you found it, and did you find it at all? https://emberjs.com/ (edited)

> about 10 seconds. I did kind of expect to see it in the community dropdown from the top menu, then looked around the main page, before I finally clicked on the "Ember community" and read the first paragraph to find it.

> In a directed search, couldn’t find it from the homepage. Looked under Community, then About, then Docs menus. Scrolled to the bottom of the page looking for links there. Finally, used the search box to search for “slack” and found the slack link buried at the bottom of the Guides intro. Searches for “discussion” and “board” turned up nothing.

> Found it right away by going to the first community link and reading

> It took me about a minute because I was skimming pages to look for Slack (rather than actually reading the text). I did find it on the Community page my second try and actually paying attention the the text. I was expecting to see the Slack icon to spot it. At a glance, the text under “Get Help from the Community” kind of blends together (sorry, I’m a skimmer when I’m looking for something specific!)

> I was expecting to see the Slack icon to spot it

> yeah, i was also looking for the slack logo

> If I didn't have that original guidance that a discussion board or Slack community existed, I'd've taken longer than the 10 seconds it took. It'd be great to see those clearly listed in the footer.
> Contributing factors to my trouble in finding the links:
> 1. The phrase "The Ember Community" placed right under the "Community" dropdown made me blink twice. The verbiage seemed redundant.
[...]

PR #3177, opened 2018-02-01
Add discussion forum to community navbar dropdown

Dog2puppy:
This pull request adds a link to the discussion forum under the community dropdown.

### DIFF
--- a/source/_header.html.erb
+++ b/source/_header.html.erb
@@ -40,6 +40,7 @@
       <li><a href="/community">The Ember Community</a>
        <li><a href="/guidelines">Guidelines</a></li>
        <li><a href="http://github.com/emberjs/">Contribute (Github)</a></li>
+       <li><a href="https://discuss.emberjs.com/">Ember Disscussion Forum</a></li>
        <li role="separator" class="divider"></li>
        <li><a href="/community/meetups">Meetups</a></li>
        <li><a href="http://jobs.emberjs.com/">Job Board</a></li>


source/_header.html.erb
```diff
@@ -40,6 +40,7 @@
       <li><a href="/community">The Ember Community</a>
        <li><a href="/guidelines">Guidelines</a></li>
        <li><a href="http://github.com/emberjs/">Contribute (Github)</a></li>
+       <li><a href="https://discuss.emberjs.com/">Ember Discussion Forum</a></li>
```
MelSumner: Thank you for the PR! 👍

I think if we add a Forum link we should probably also add the Slack community link, but then again, both are front and center on the community page, which is already in this dropdown. @jenweber @acorncom any thoughts?

acorncom: @MelSumner inclined to agree with you. While having both links easily findable in the menu would be handy, we need to curate the items in this list so we can help people clearly find major areas of the community. I'm inclined to say we don't want to merge this. @jenweber @locks ?

jenweber: I am kinda torn. On one hand, I don't want to clutter the nav. On the other hand, new users need those links the most and I don't know that they would find them as-is. I would like to do an informal UX test before making a decision. I'll show some new devs our site, and ask them to find the discussion board/slack chat.

Dog2puppy: I just thought it was a good idea because the discussion forum seemed hidden away.

MelSumner: @Dog2puppy do you think that if we simplified the Community page in general, it would be clearer for you? See #3156 & see what you think.

PR #3180, opened 2018-02-08
Feature/builds style prep

MelSumner:
- added in some styles for an inline nav/button to help navigation for builds app


MelSumner:
maybe related to https://github.com/ember-learn/builds/pull/67

PR #3149, opened 2018-01-11
Ember 3.0 Release Blog Post

locks:
- [Rendered](https://github.com/emberjs/website/blob/3-0-release-blog-post/source/blog/2018-02-14-ember-3-0-released.md)
- Preview
- 3.1-beta blog post: https://github.com/emberjs/website/pull/3187

## Sign-offs

- [ ] Ember.js (@chancancode)
- [ ] Ember Data (@bmac)
- [ ] Ember CLI (@rwjblue)
- [ ] Website (@locks, @mixonic)

## Checklist

- Ember
  - [ ] Blog
  - [ ] v3.0.0
  - [ ] v3.1.0-beta.1
- Ember Data
  - [ ] Blog
  - [ ] v3.0.0
  - [ ] v3.1.0-beta.1
- Ember CLI
  - [ ] Blog
  - [ ] v3.0.0
  - [ ] v3.1.0-beta.1

See https://github.com/emberjs/website/pull/2824/files for inspiration.

PR #3184, opened 2018-02-09
Make clear what Ember 3.0 release means to globals mode users

Gaurav0:


source/blog/2018-02-12-ember-3-0-released.md
```diff
@@ -0,0 +1,250 @@
...
+
+You can read more about our detailed transition plans through Ember 3.5
+in
+[The Road to Ember 3.0](https://www.emberjs.com/blog/2017/10/03/the-road-to-ember-3-0.html)
+and below.
+
+## API Removals in 3.0
+
+All APIs to be removed in Ember 3.0 are already deprecated today. By removing support for these features we achieve two goals for our major release. First, we make it simpler for Ember to strip code that supports these features if you don’t use them. Second, we make it possible for Ember to remove the code supporting these features entirely, opening the door to refactoring that improves the clarity and simplicity of our implementations.
+<hr>
```
Gaurav0: Above copied from 2.18 blog post

source/blog/2018-02-12-ember-3-0-released.md
```diff
@@ -0,0 +1,250 @@
...
+
+In Ember 3.0 the `Ember.K` function will be removed. For more details, please see the deprecation guide:
+
+- [`Ember.K`](/deprecations/v2.x/#toc_code-ember-k-code)
+
+#### Consuming Ember through Bower
+
+Ember.js 3.0 and later releases will no longer be published to bower. If you still use bower today, we strongly suggest you begin the migration of your dependencies to npm.
+
+<hr>
```
Gaurav0: All of the above from previous comment copied from Road to Ember 3.0 with minor removals for brevity.

source/blog/2018-02-12-ember-3-0-released.md
```diff
@@ -0,0 +1,250 @@
...
+
+In the past, we have failed to make clear what this means for users of globals mode. We apologize for this lack of clarity.
+
+In short, globals builds of any kind will no longer be maintained or published. Not only will they not be availabe in bower, but you
+will likely not find them on CDNs or by download from the emberjs.com website. While the initial release of Ember 3.0 does contain
+a globals build of bower as part of the ember-source npm package, this too will disappear before long. What this means is,
+if you need to continue using globals mode in Ember 3.0, as a part of your build process, you will need to compile a build of Ember itself
+using ember-cli. We know this will cause considerable hardship to users of globals mode, but this is important in order to be able to make
+progress on some long awaited features such as splitting Ember into smaller packages and tree shaking.
```
Gaurav0: Above from previous comment should be given close attention.

Remainder below is unaltered.

acorncom: @Gaurav0 I would personally think the handwriting has been on the wall for globals mode for a while, but that may be because I've been relatively close to things. Maybe we could do something along these lines?

> Continuing our shift away from bundling Ember as one large Javascript file, which we have been aiming for since the introduction of Ember-CLI, we’re now officially ending the publishing of new Ember versions on Bower or CDNs.
> 
> We recognize that for a small part of our community, this may require reworking build processes (it will entail additional work for Ember Twiddle as well). We're sorry about that, but feel unlocking long-awaited features such as tree-shaking and allowing publishing Ember as smaller packages makes this a worthwhile cost to pay as a community.

The above probably needs further wordsmithing, but I think conveys the goals behind these changes ... @rwjblue @wycats @locks any thoughts on the above? I wondered about talking about the goal of allowing easier npm access in other build systems, but wasn't sure if that'd be promising too much ...

Gaurav0: Hi, @acorncom. Thanks for working on the wordsmithing. I appreciate the feedback and I'm incorporating it. I'm also going to make the file rename to correct the date a separate commit.

Gaurav0: I would personally think the handwriting has been on the wall for globals mode for a while

Perhaps, which is why I am encouraging us to deprecate it explicitly, and soon. It is important to communicate clearly and explicitly when something is going away, as soon as we know it is, and then give people lots of time to migrate.

My concern that brought up this PR is that I didn't translate automatically from "publishing to bower is going away" to "all builds (globals or otherwise) are going away and I need to compile Ember myself using ember-cli". In fact, several other core team members I spoke to weren't aware of this either.

Similarly, if using globals mode now is not recommended because it might be going away, well that is kinda the definition of the word deprecated. People get upset when they aren't warned, so let's at least avoid the next mistake if we can't fix this one.

mixonic:
Thanks @Gaurav0 I'm going to merge this and take it as a starting point. I'm sure there will be pretty heavy alterations to review on https://github.com/emberjs/website/pull/3149. 👍

PR #3192, opened 2018-02-15
Point to new deprecations app

sivakumar-kailasam:


sivakumar-kailasam:
@locks If we don't care about the subpath level redirects like `emberjs.com/guides/routing` anymore its a one liner change in static.json and we can ship this even with existing apps.

PR #3197, opened 2018-02-18
RFC#297 | Ember.Logger deprecation

lupestro:
This is the deprecation text to go with RFC 297 ("Deprecation of Ember.Logger") currently a PR on emberjs/ember.js. I'm entertaining feedback on the text now, but it shouldn't be merged until that PR is also merged.

### DIFF
--- a/source/deprecations/v3.x.html.md
+++ b/source/deprecations/v3.x.html.md
@@ -54,41 +54,48 @@ object.notifyPropertyChange('someProperty');
 
 Calling `array.get('@each')` is deprecated. `@each` may only be used as dependency key.
 
+### Deprecations Added in Upcoming Versions
+
 #### Use console rather than Ember.Logger
 
-##### until: 3.5.0
+##### until: 4.0.0
 ##### id: ember-console.deprecate-logger
 
-The `Ember.Logger` module will be removed after the first LTS of the 3.x cycle. You should remove any calls to `Ember.Logger` and use calls to `console` instead.
+The `Ember.Logger` module will be removed at the start of the 4.x cycle. You should remove any calls to `Ember.Logger` and use calls to `console` instead.
 
 In Edge and IE11, uses of `console` beyond calling its methods may require more subtle changes than simply substituting `console` wherever `Logger` appears. In these browsers, they will behave just as they do in other browsers when your development tools window is open. However, when run normally, calls to its methods must not be bound to anything other than the console object or you will receive an `Invalid calling object` exception. This is a known inconsistency with these browsers. (Edge issue #14495220.)
 
-For example, the following uses will not work in Edge and IE11 except when the development tools window is open:
+To avoid this, transform the following usage:
 
 ``` javascript
-var print = console.log; // print is unbound, therefore bound to global
-print(message);
-
-// Using anonymous unbound local
-(console.debug ? console.debug : console.log)(message);
-
-//Explicitly binding to nothing or to something else
-console.info.apply(undefined, arguments);
-console.info.apply(null, arguments);
-console.info.apply(this, arguments);
+var print = Logger.log; // assigning method to variable
+```
+to:
+``` javascript
+var print = console.log.bind(console); // assigning method bound to console to variable
 ```
 
-Instead, use the following, which will work consistently across browsers:
-
+Also, transform any of the following:
 ``` javascript
-var print = console.log.bind(console);
-print(message);
-
-if (console.debug) {
-  console.debug(message);
-} else {
-  console.log(message);
-}
-
+Logger.info.apply(undefined, arguments); // or
+Logger.info.apply(null, arguments); // or
+Logger.info.apply(this, arguments); // or
+```
+to:
+``` javascript
 console.info.apply(console, arguments);
 ```
+
+Finally, because node versions before version 9 don't support console.debug, you may want to transform the following:
+``` javascript
+Logger.debug(message);
+```
+to:
+``` javascript
+if (console.debug) {
+console.debug(message);
+} else {
+console.log(message);
+}
+```
+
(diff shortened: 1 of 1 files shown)


source/deprecations/v3.x.html.md
```diff
@@ -53,3 +53,42 @@ object.notifyPropertyChange('someProperty');
 ##### id: ember-metal.getting-each
 
 Calling `array.get('@each')` is deprecated. `@each` may only be used as dependency key.
+
+#### Use console rather than Ember.Logger
```
locks: This deprecation should be under "upcoming deprecation" as it is not included in a release as of yet :)

source/deprecations/v3.x.html.md
```diff
@@ -53,3 +53,42 @@ object.notifyPropertyChange('someProperty');
...
+##### id: ember-console.deprecate-logger
+
+The `Ember.Logger` module will be removed after the first LTS of the 3.x cycle. You should remove any calls to `Ember.Logger` and use calls to `console` instead.
+
+In Edge and IE11, uses of `console` beyond calling its methods may require more subtle changes than simply substituting `console` wherever `Logger` appears. In these browsers, they will behave just as they do in other browsers when your development tools window is open. However, when run normally, calls to its methods must not be bound to anything other than the console object or you will receive an `Invalid calling object` exception. This is a known inconsistency with these browsers. (Edge issue #14495220.)
+
+For example, the following uses will not work in Edge and IE11 except when the development tools window is open:
+
+``` javascript
+var print = console.log; // print is unbound, therefore bound to global
```
locks: Why is this mentioned? Seems unrelated to the deprecation.

lupestro: When the user changes their code to silence the deprecation, it will at first appear that a simple text substitution of “Logger”=>”console” will do. However, the moment anybody runs their code with Edge or IE11 *without the debugger open*, they’ll face-plant into this issue. The solution is simple but non-obvious, so I figured I’d better document it so anybody who is updating their code by hand will be equipped to handle it as they make their change. This is something the codemod will also need to deal with or flag for manual intervention.

source/deprecations/v3.x.html.md
```diff
@@ -53,3 +53,57 @@ object.notifyPropertyChange('someProperty');
...
+``` javascript
+Logger.info.apply(undefined, arguments); // or
+Logger.info.apply(null, arguments); // or
+Logger.info.apply(this, arguments); // or
+```
+
+to:
+
+``` javascript
+console.info.apply(console, arguments);
```
locks: Thoughts on `console.info(...arguments)`?

sivakumar-kailasam:
@lupestro We're migrating deprecations from this repo into its [own app](https://github.com/ember-learn/deprecation-app/). Can you please raise a new PR following these [instructions](https://github.com/ember-learn/deprecation-app/blob/master/README.md#adding-new-deprecations)?

lupestro:
@sivakumar-kailasam [https://github.com/ember-learn/deprecation-app/pull/91]()

sivakumar-kailasam:
Thanks @lupestro closing this in favor of the new PR.

source/deprecations/v3.x.html.md
```diff
@@ -53,3 +53,57 @@ object.notifyPropertyChange('someProperty');
 ##### id: ember-metal.getting-each
 
 Calling `array.get('@each')` is deprecated. `@each` may only be used as dependency key.
+
+### Deprecations Added in Upcoming Versions
```
rwjblue: Sorry for the conflicting advice, can you change this to be within the 3.2 section? I'm pretty sure we will be able to tackle it before 3.2 goes to beta...

source/deprecations/v3.x.html.md
```diff
@@ -53,3 +53,57 @@ object.notifyPropertyChange('someProperty');
...
+
+#### Use console rather than Ember.Logger
+
+##### until: 4.0.0
+##### id: ember-console.deprecate-logger
+
+The `Ember.Logger` module will be removed at the start of the 4.x cycle. You should remove any calls to `Ember.Logger` and use calls to `console` instead.
+
+In Edge and IE11, uses of `console` beyond calling its methods may require more subtle changes than simply substituting `console` wherever `Logger` appears. In these browsers, they will behave just as they do in other browsers when your development tools window is open. However, when run normally, calls to its methods must not be bound to anything other than the console object or you will receive an `Invalid calling object` exception. This is a known inconsistency with these browsers. (Edge issue #14495220.)
```
rwjblue: The various caveats should be under a specific section (e.g. `Migration Gotchas` or somesuch). I see two main gotchas here:

* calling console methods unbound 
* console.debug being missing

FWIW, I think the specific guidance for the latter item there is to just not use console.debug...

source/deprecations/v3.x.html.md
```diff
@@ -53,3 +53,57 @@ object.notifyPropertyChange('someProperty');
...
 Calling `array.get('@each')` is deprecated. `@each` may only be used as dependency key.
+
+### Deprecations Added in Upcoming Versions
+
+#### Use console rather than Ember.Logger
+
+##### until: 4.0.0
+##### id: ember-console.deprecate-logger
+
+The `Ember.Logger` module will be removed at the start of the 4.x cycle. You should remove any calls to `Ember.Logger` and use calls to `console` instead.
```
rwjblue: Directly under this line, can you show a few before and after examples?

I'm thinking something like:


```js
// Before
Ember.Logger.log('my little pony');

// After
console.log('my little pony');
```

source/deprecations/v3.x.html.md
lupestro: The section headings at this level aren't terribly distinct, but I broke it up under headings. Looking at a lot of the other deprecations in 3.x and 2.x, the description for this one is a bit more involved, but I think the detail is necessary.

source/deprecations/v3.x.html.md
lupestro: Added. While I was at it, I combined the before/after pairs down on the other examples.  Hopefully that doesn't violate editorial conventions too badly. They do read better with the code lines closer together and, with three of them, it saves significant space and plays down the changes a little.

lupestro:
When the editing is complete and this is merged, I'll migrate the content over to deprecation-app as well, so it's there for the future.

PR #3200, opened 2018-02-21
Remove old guides content & replace with nginx redirect

sivakumar-kailasam:
Couple of blockers before we can go down this path: 

* Heroku buildpack doesn't support wildcard redirects along with named redirects yet but there's a fork out there which works. 
* The new guides app needs to go live so that we can point to that for `/release` paths


locks:
@sivakumar-kailasam we have a way to get the latest version of Ember.js. Could we do that while the Guides app isn't production-ready?

sivakumar-kailasam:
@locks do we care about the subpath redirects (like emberjs.com/guides/routing)? if so we'll have to wait until the ember app lands along with proxy changes. If we don't we can redirect to guides and it does load latest by default.

The existing implementation on the website seems to be the only alternative afaik. If you had other ideas let me know.

locks:
@sivakumar-kailasam at this point, I think it's fine to redirect to the root of the Guides. we haven't had `/guides` in like 4 years?

sivakumar-kailasam:
Yeah feels like forever. I've rebased & corrected the commits to redirect to guides.

sivakumar-kailasam:
@locks Good to merge

PR #3203, opened 2018-02-21
blog posts. again.

MelSumner:
- added 'responsive: true' to more blog posts (including the most current) so they are device-agnostic
- updated some of the page .scss files to include the base file import that they were missing. 

(will do the button component updates in a different PR)


acorncom:
Is this needed with the change that occurred in https://github.com/emberjs/website/pull/3189 ?

MelSumner:
@acorncom that doesn't seem to have worked- when I checked the survey link on my phone, it still was the non-responsive version. I don't have time to troubleshoot the permanent solution right now, so we'll just have to make sure that blogs have the 'responsive: true' line in them.

PR #3204, opened 2018-02-22
some style updates

MelSumner:
- made the max page width consistent with other elements
- made the buttons consistent on the blog and survey page
- fixed the a11y of the button with the image
- put a tinge of color on the body background- the white is just toooooo stark. 
- made the background of the survey reminder the same accent blue used for the "jobs" image and the post author names (we can ditch this if @mixonic doesn't like it)

Screenshot of using the blue on the survey banner: 
image

### DIFF
--- a/source/stylesheets/components/buttons.css.scss
+++ b/source/stylesheets/components/buttons.css.scss
@@ -1,48 +1,34 @@
 @import 'bourbon';
 
 .button {
+  background-color: $orange;
+  border-radius: 5px;
+  box-shadow: 0 1px 0 rgba(255,255,255,0.35);
+  color: $white;
   display: block;
-  width: 220px;
-  height: 40px;
+  font-size: 0.9em;
   font-weight: bold;
-  text-transform: uppercase;
-  text-decoration: none;
-  line-height: 38px;
-  font-size: 14px;
-  color: #f6ece8;
+  height: 40px;
+  letter-spacing: 1px;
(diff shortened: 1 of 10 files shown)


mixonic:
@MelSumner Seems great. Blue bar seems fine- pops great on desktop.

The button on the landing page animates nicely on hover: https://ember-website-staging-pr-3204.herokuapp.com/ember-community-survey-2018/

The one on the blog post just gets taller: https://ember-website-staging-pr-3204.herokuapp.com/blog/2018/02/21/2018-ember-community-survey.html

Should they be the same?

source/stylesheets/components/buttons.css.scss
```diff
@@ -65,13 +52,13 @@
   text-transform: uppercase;
   text-shadow: rgba(0, 0, 0, 0.3) 0 1px 0;
 
-  &.orange {
-    background-image: url('/images/small-orange-button.png');
-  }
+  // &.orange {
```
mixonic: commented stuff can probably be removed.

MelSumner:
oh and it turns blue! def not supposed to do that. I'll make sure the hover transitions are the same all the way around.

MelSumner:
Because both the job board banner and the survey banner are using the same background color, and the job board is now included in the site navigation, the job banner is turned off on mobile devices until after EmberConf 2018. Issue #3205 will track this.

PR #3211, opened 2018-03-02
Add deprecation for Route#router

cibernox:
I initially added it two 3.1 thinking that the deprecation itself can be considered a [BUGFIX beta], but it's not confirmed.


jenweber:
@cibernox we just shipped https://github.com/ember-learn/deprecation-app as the new home for deprecations. Could you PR it there? Thank you!

rwjblue:
Also discussed on the implementation PR that this should target 3.2

PR #3213, opened 2018-03-04
Update _header.html.erb

locks:
Avoid ugly redirect.


jenweber:
works just as well in production

PR #3219, opened 2018-03-08
Add Fitbit logo

GCheung55:
Adding Fitbit's logo as a company that uses Ember.js.


locks:
Thank you for submitting @GCheung55 :) Could you update the image to be 420x206, so it matches the other "hi-res" logos we have?

Issue #3217, opened 2018-03-08
Make community chat resources a list instead of paragraph


jenweber:
Right now, a user has to read a paragraph of text on [this page](https://emberjs.com/community/) to find out which forums exist for Q&A. Instead, this should be a list of resources with the name and then a description. Stack Overflow should be listed last. If you're feeling fancy, you could use icons for each service in place of bullets list, but please do not import any new fonts. SVGs/images are ok.

How things look today:
![screen shot 2018-03-07 at 11 17 25 pm](https://user-images.githubusercontent.com/16627268/37132834-e15962da-225d-11e8-8088-2de94ab17223.png)

Note, this issue is set aside for the EmberConf contributors workshop on 3/12. If it's not claimed by the end of that day, feel free to grab it!

sivakumar-kailasam:
React website's uses text base links to all their channels in the footer. It probably makes it easy to discover slack as well as discuss.emberjs.com if we adapt it. Thoughts?

<img width="234" alt="screenshot 2018-03-08 09 59 33" src="https://user-images.githubusercontent.com/604117/37133435-603942ae-22b9-11e8-861e-b8254de5b8c1.png">

bagby:
I'll take a crack at this.

jenweber:
Closed by merging #3220

PR #3220, opened 2018-03-12
Clean up Community Help Links

bagby:
First pass at fixing #3217 . This just cleans up that section and makes it more apparent where the help is.

image

### DIFF
--- a/source/community.html.erb
+++ b/source/community.html.erb
@@ -18,20 +18,25 @@ responsive: true
 
   <div class="image"></div>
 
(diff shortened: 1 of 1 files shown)


jenweber:
@bagby could you remove the references to IRC and the link? We took it out a while ago. The final sentence could be changed to something like "Please note that these resources are not managed or moderated by the Ember Core Team. They are public forums."

Thank you! After that, this looks good to merge!

bagby:
@jenweber will do, let me get off EmberConf time so I have the focus to do it :)

jenweber:
Lookin' good. Thanks @bagby for your first PR to the website! This is a much cleaner layout 🐹

PR #3227, opened 2018-03-16
Add Alex to the team

twokul:


source/team.html.erb
```diff
@@ -86,7 +86,7 @@ responsive: true
 <hr/>
 <section class="team section">
 <h2 class="text-center">
-    Ember Core Alumni
+    Ember Core Emeritus
```
acorncom: Hmm, this renames all Ember Core folks from Alumni to Emeritus. Is that what we intend across the board?

twokul: Yeah. Based on the conversations with several core team members (last one was Rob), the intention is to call Ember Core folks "emeritus" (preserve all the privileges w/o any further obligations/responsibilities). I'm happy to submit a separate PR if needed

rwjblue: Yes, Emeritus is the correct term (and is what we use internally also)

twokul:
@rwjblue @acorncom done and done.

Issue #3229, opened 2018-03-20
Remove EmberConf callout


locks:
As this year's EmberConf has come and go, so should the banner on the homepage.
Removing [this code](https://github.com/emberjs/website/blob/master/source/about.html.erb#L53-L62) should be enough.

locks:
Addressed by https://github.com/emberjs/website/pull/3234.

PR #3234, opened 2018-03-22
Remove EmberConf callout #3229

urbany:
As requested in #3229


locks:
First PR en route to CI and then the deployed website!

PR #3255, opened 2018-03-31
clean up messaging around deprecations on the learn page

jenweber:
We can do better from a marketing perspective than this:

> Over the course of the years cruft inevitably sneaks into all software projects. 

Proposed changes:
screen shot 2018-03-31 at 9 11 26 am

I removed the extended links to different versions. With the new app and nicer landing page, I don't think they're necessary.

Sister PR in the deprecations app https://github.com/ember-learn/deprecation-app/pull/112

### DIFF
--- a/source/learn.html.erb
+++ b/source/learn.html.erb
@@ -37,32 +37,14 @@ responsive: true
     <section>
       <h2 id="deprecations" class="anchorable-toc"> <a href="http://emberjs.com/deprecations/">Deprecation Guides<i class="icon-link-ext"></i></a></h2>
       <p>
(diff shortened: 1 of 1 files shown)


MelSumner:
What if we included something about codemods? (maybe @rwjblue would have a comment about how those are an integral part of the deprecation process?)

jenweber:
@MelSumner done. I'm repeating myself a bit between the FAQ, this, and the deprecation guide, but I don't think that's a bad thing. See https://github.com/ember-learn/deprecation-app/pull/112

MelSumner:
I agree @jenweber - it's not a bad thing. :+1:

PR #3249, opened 2018-03-30
update to the teams page

MelSumner:
**Done:**
- updated the data so names can be sorted by last name
- separated teams into section
- removed subteams reference, replaced it with team name headings
- added Github and twitter to everyone
- removed team badges (team headings take care of it)

**Still to do:** 
- [ ] need twitter handles for everyone (or plan for what to do in lieu of twitter handles)
- [ ] need descriptor sentences for each team- 1-2 sentences that explain what each team is primarily responsible for.  I'm going to tag a few people in each team- please weigh in on the comments if you have a suggestion.  
  - Ember CLI (@kategengler @kellyselden @Turbo87) 
  - Ember Data (@dgeb @igorT)
  - Learning Team (@locks @jenweber @acorncom)
- [ ] need details for each team (repo for the public meeting notes, when do they meet, etc)

PR #3257, opened 2018-04-01
typography: more spacing below h1 and h2

jenweber:
This is a tiny fix for something that bugs me - I think heading text looks more balanced when it has greater spacing than a typical line. Let me know what you think. I had a designer look through and he said that the spacing I'm adding should not be applied to other h2s throughout the site, just the home page, which is what I've done here.

### After:
screen shot 2018-04-01 at 7 36 47 pm

screen shot 2018-04-01 at 7 36 39 pm

### Before
screen shot 2018-04-01 at 7 37 29 pm

screen shot 2018-04-01 at 7 37 22 pm

### DIFF
--- a/source/stylesheets/about.css.scss
+++ b/source/stylesheets/about.css.scss
@@ -93,6 +93,11 @@ body.index {
   background: no-repeat top center #faf4f1;
   background-image: url("/images/background-shades.png");
   background-image: url('/images/background-shades.svg'), none; // See: http://css-tricks.com/svg-fallbacks/
+  
+  h2 {
+    padding-bottom: 0.5em;
+  }
+
   @media screen and (min-width: 2000px) {
     // Stretch background width when wider than source image
     background-size: 100% auto;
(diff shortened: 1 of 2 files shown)


MelSumner:
@jenweber it's already in PR #3249

wifelette:
While you're in here, are we also able to fix the business of the logos being skewed to the bottom in the carousel?

MelSumner:
@wifelette fixed with https://github.com/emberjs/website/pull/3266

jenweber:
@MelSumner done! However by putting this rule in base_typography, it now affects all h2s, while I was initially targeting only the home page. Have a look and see if this is an improvement. It especially has an impact on the blog headings (list view and individual view). Thanks!

MelSumner:
I'd rather add a margin-bottom to the h2 across the whole site and remove the hr. We can switch to a border bottom (or just leave the line out entirely) for the blog; I'm trying to avoid doing custom implementations to each page, since this is why the CSS is a mess already and felt like we extracted the typography styles (across the board) out as a first step. What do you think?

MelSumner:
@jenweber since we mostly resolved this issue in other PRs, I'm going to close this PR. please re-open if you don't feel like the issue was adequately addressed. 👍

PR #3263, opened 2018-04-05
resolved issue where things were too scrunched on mobile devices

MelSumner:
- @sivakumar-kailasam can you check the deploy to see if this resolves the mobile issue you observed
- need to still cross-browser check this PR to make sure it's okay in IE


sivakumar-kailasam:
I tried it on IE edge & mobile and seems its good on both. Should the twitter & github links open in a new tab instead of opening up in the current tab?

MelSumner:
@sivakumar-kailasam nope- not ideal for a11y or security.

PR #3279, opened 2018-04-14
Fix date of 3.1 blog post

ryanto:
Update the date of the 3.1 blog post and add a redirect so the old url still works


Gemfile.lock
```diff
@@ -170,16 +170,6 @@ GEM
       coderay (~> 1.0)
       method_source (~> 0.8)
       slop (~> 3.4)
-    pry (0.9.12.6-i386-mingw32)
```
MelSumner: Can you help me understand why you removed these?

ryanto: this was the automatic result of a fresh bundle install.

MelSumner: okay. Let me pull down this branch first and check that it won't break a windows build :)

ryanto: 👍 if it does I can figure out how to get this into the Gemfile so anyone who starts fresh will end up with the correct dependencies. Keep me posted!

PR #3280, opened 2018-04-14
[work in progress/experiment - do not merge]

MelSumner:
Trying to see if we like the blog when it has visible archives. 
- adding tags to blog posts 
- adding additional categories to sidebar


MelSumner:
@sivakumar-kailasam thanks for catching that, I needed to update the tab.html.erb to more accurately reflect current design. :+1:

PR #3287, opened 2018-04-16
Remove Ember CLI instructions from the call to action.

locks:
The Ember CLI instructions are only useful if you already
have enough context to understand what they accomplish.
To that end, I think directing people to our quick start guide
should be enough.


MelSumner:
I disagree with this one.

PR #3304, opened 2018-04-19
Add Ember Learn update for Ember Times 43 or 44

amyrlam:
UPDATE: Feel free to pass on this for the April 20 edition if there is not time, it can go in next week also!

Never mind, unstaged it:
~~Aside: it seems like my `.editorconfig` is making some whitespace changes to copy I didn't touch as well. Can fix if undesired.~~


amyrlam:
For: https://github.com/emberjs/website/pull/3303 or next week's Ember Times

jessica-jordan:
That's a great write up, thank you for getting that together!

PR #3306, opened 2018-04-19
temporarily removing API search from the website

mansona:
For anyone who doesn't know the context of this change: as discussed on today's learning team meeting we need to temporarily disable API search results on the website because of issues with our index. There are separate efforts trying to fix that and we can re-enable API search on the website if any of those efforts pay off.


mansona:
@sivakumar-kailasam I'm not sure that's technically true 🤔 https://cl.ly/2Z3p1x1B1J3J

I guess this is one of those things where it might depend on what you're testing in the search i.e. maybe packages don't work but function names do or something?

sivakumar-kailasam:
Sure this makes sense to me now, I'll follow up with algolia to see how we can mix this up with the custom index we've built for api docs.

Issue #3137, opened 2018-01-07
Add Invitation to Contribute to page layout


MelSumner:
Goal: to add a "Contribute to this page" link to each page. 

Criteria: 
- must be at the bottom of each page, centered, above the footer
- must link to: https://github.com/emberjs/website 
- should be responsive (device-agnostic)
- should be accessible 
- should complement the existing site design

Guidelines:   
- code should go into the layout.erb file
- css should go into the site.css.scss file (at the bottom of the file is fine)

SeanHealy33:
I've had a quick look at implementing this @MelSumner 

![screenshot from 2018-04-04 12-10-09](https://user-images.githubusercontent.com/6124835/38304386-25aedf60-3801-11e8-8530-7e964b73f85b.png)

Any thoughts on this attempt? It feels like it's a too close to the gear/donate section.

MelSumner:
Thanks for doing this!

PR #3313, opened 2018-04-23
Add a link to contribute on the bottom of all pages

SeanHealy33:
Relates To: https://github.com/emberjs/website/issues/3137

Hover:
screenshot from 2018-04-23 19-02-36
Focused:
screenshot from 2018-04-23 19-23-43

### DIFF
--- a/source/layout.erb
+++ b/source/layout.erb
@@ -52,6 +52,11 @@
       <div id="content" class="main<% if sidebar %> has-sidebar<% end %>" role="main">
         <%= yield %>
       </div>
+      <div class="cta-contribute">
+        <a title="Use GitHub to suggest an edit to this page" href="https://www.github.com/emberjs/website">
+          Contribute to this page
+        </a>
+      </div>
     </div>
 
     <main-footer></main-footer>
--- a/source/stylesheets/base/_typography.scss
+++ b/source/stylesheets/base/_typography.scss
@@ -12,7 +12,7 @@ body {
 a {
   text-decoration: none;
   color: $orange-darker;
-  &:hover, :focus {
+  &:hover, &:focus {
     color: $orange-darkest;
   }
 }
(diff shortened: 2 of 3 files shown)


source/stylesheets/base/_typography.scss
```diff
@@ -12,15 +12,15 @@ body {
...
+  &:hover, &:focus {
     color: $orange-darkest;
   }
 }
 p {
   a {
     color: $orange-darker;
     text-decoration: underline;
-    &:hover, :focus {
+    &:hover, &:focus {
```
SeanHealy33: This was pretty interesting, the focus selector here was unreachable since the css attribute looked like `a :focus` instead of `a:focus` this meant the bootstrap `a:focus` was being used to style the link blue

source/layout.erb
```diff
@@ -52,6 +52,11 @@
       <div id="content" class="main<% if sidebar %> has-sidebar<% end %>" role="main">
         <%= yield %>
       </div>
+      <div class="cta-contribute">
+        <a title="Use GitHub to suggest an edit to this page" href="https://www.github.com/emberjs/website">
```
kennethlarsen: Would it be a good idea to have this open in a new tab with `target= _blank`?

SeanHealy33: I like this idea. Want me to update the pr to include `target='_blank'`  @kennethlarsen?

MelSumner: If you do decide you want to go this route, please consider a new PR, ensuring that the solution you implement follows WCAG Guideline 3.2 (Predictable) and perhaps implements technique [G201: Giving users advance warning when opening a new window](https://www.w3.org/TR/WCAG20-TECHS/G201.html). Thank you! :+1:

MelSumner:
Thank you for doing this @SeanHealy33 :+1:

PR #3329, opened 2018-05-06
Update blog link for Ember Times

amyrlam:
Original Twitch link expired after 2wk. Jen updated it in her Medium post as well. re: https://embercommunity.slack.com/archives/C04KG57CF/p1525524488000007?thread_ts=1525470060.000346&cid=C04KG57CF

@jenweber @jessica-jordan


amyrlam:
I don't think we want this cause @sivakumar-kailasam mentioned it would mess with RSS, so closing.

Maybe could include the right link next issue.

Issue #3273, opened 2018-04-12
Hover issue social links in footer


anpenava:
I think text decoration just needs to be none, I can make PR tonight if nobody have time for styling issues :D

`.footer .footer-social a:hover {
  text-decoration: none;
}`

MelSumner:
@anpenava can you tell me what page you're seeing this issue on? I thought we already fixed it. Thanks! :+1:

anpenava:
I am using most recent google chrome browser on macbook, I added screenshot.
[Link to screenshot](https://ibb.co/hGqLz7)

anpenava:
@MelSumner sorry on late response, now I can see bug only on this page 
 https://guides.emberjs.com/v3.1.0/

Willibaur:
@MelSumner  :point_up:  PR is in place for solving this issue

jenweber:
Closed, addressed by PRs to ember-styleguide. Thanks everyone for reporting!

PR #3341, opened 2018-05-16
Hover issue social links in footer #3273

ashokaditya:
Although, I can't verify the patch locally cause localhost:4576/guides redirects me to the production https://guides.emberjs.com/v3.1.0/

### DIFF
--- a/source/stylesheets/_footer.scss
+++ b/source/stylesheets/_footer.scss
@@ -149,6 +149,7 @@
 }
 
 .footer .footer-social a:hover {
+  text-decoration: none;
   color: #dd6a58;
   opacity: 1;
 }


jenweber:
@ashokaditya can you tell me more about the problem this tries to fix? I don't see any issues with the hover state on the social links.

ashokaditya:
@jenweber The social links at the bottom of https://guides.emberjs.com/v3.1.0/ show underlines when hovered which is inconsistent with the rest of the website's "no underline" on hover. My patch is to fix that. But like I mention I can't verify the fix locally cause the localhost:4567/guides linke redirects me to https://guides.emberjs.com/v3.1.0/.

Please let me know if that clarifies it :)

Thanks.

jenweber:
Oh ok, I see it now. I was looking on the Ember home page and not the guides, oops. We're launching a new app for the Guides this week (ish) and the hover issue is already fixed there. I'm going to close this PR since it's covered elsewhere, but thank you for taking the time to report/fix the issue!

ashokaditya:
👍

Issue #3346, opened 2018-05-17
Who's Using Ember.js? slider - logos jump down when changing between last and first slide


thisFunction:
When changing between last and first slide the logos jump down.

[Click here to see video](https://photos.google.com/share/AF1QipOTVGDHXNbC_IXDIA8L876swhUvmZKdQTvcV9Heav9GxvCZIs50UJS5K-s59faNRw?key=WTQ4VE9YWEtBaGI4bmJmOVFmVXdYZWNXaHBFWTRR)

MelSumner:
This is a duplicate of https://github.com/emberjs/website/issues/3134 - thank you! :+1:

PR #3347, opened 2018-05-17
[#3346] Remove front page slider jump behavior

Martodox:
Sample video:
https://photos.app.goo.gl/JjD0qQn7sRSumvVD2

Previous version:
https://photos.app.goo.gl/IUZakNzQBMDc84qv1


Martodox:
Sample video:
https://photos.app.goo.gl/7x1tRZNm026UmqzS2

Martodox:
@jenweber it is a definite improvement over previous version. I think in order to make it look good that jquery plugin that is being used as a slider component needs to be removed/replaced with something modern, responsive? What do you think?

MelSumner:
Hey all- this is sort of a two fold thing- 

1. there's a newer version of the slider we use, could update (has side effects)
2. as we're working on Emberizing the website (see [ember-styleguide](http://github.com/ember-learn/ember-styleguide)) there's sort of the question as to whether or not updating this will be worth it, or if we should push forward with the replacement. 

@jenweber thoughts?

jenweber:
If there are no objections, I think this PR is a net improvement and we should merge it, then worry about other upgrades in later PRs

MelSumner:
Could we make sure the other viewports work too? Here's what I see for a small viewport: 
image

jenweber:
closed with merging https://github.com/emberjs/website/pull/3648

Thanks for trying to help us fix this thing! Multiple people made attempts and the carousel itself was so buggy that a recent contributor just made it a static box.

PR #3359, opened 2018-05-23
Home page survey for redesign

ryanto:
This blog post is a draft of a small survey we'd like to send to Ember users. Hopefully, this will let us find out which of Ember's values resonate most with the community, and how we can use those values to one day redesign the homepage.

Also this PR can help communicate to core & friends just exactly what we're doing with the survey. Let's make sure folks aren't caught by surprise since there's already an existing Ember survey. ❤️


ryanto:
Here's a screenshot so folks can quickly get an idea of this...

localhost_4567_blog_2018_05_28_ember-homepage-survey html 1

wifelette:
Added this to the agenda for the Core Team Meeting Friday :)

MelSumner:
@ryanto what else would you need to move this forward?

ryanto:
Thanks for reminding me @MelSumner !

ryanto:
Heya! I just pushed some code to update the date.

MelSumner:
@ryanto since the date is in the future, it doesn't show up in the deploy. Can you update the date to today's date please?

ryanto:
Opps! I totally missed this. If we deploy again will it show up?

Issue #47, opened 2012-02-04
Need Documentation on Ember.StateManager / Ember.State


patr1ck:
Right now there's a hole in the documentation with regards to Ember's StateManager. From what I've seen it's pretty useful, so it would be good to have docs for it.

PR #3363, opened 2018-05-25
Remove Ember.js Times #47 from Recent Posts Category

jessica-jordan:
With the new categorization feature of blog posts that has been launched through https://github.com/emberjs/website/pull/3280 the Ember.js Times now has their own dedicated post category. 

This moves the Issue No. 47 which was just launched last week away from the more general `Recent Posts` section and makes it only show up in the dedicated embertimes section instead.

| **before** | **after** |
|------------|----------|
| screen shot 2018-05-25 at 08 42 22| screen shot 2018-05-25 at 08 42 04 |


jessica-jordan:
@MelSumner Can you take a look?

PR #3368, opened 2018-05-28
spelling: fix design spelling in community blog post

rtablada:


sivakumar-kailasam:
I can't recollect if that one was intentional

jenweber:
I have seen `devsign` used as a pun by @MelSumner. I think it's probably intentional. Perhaps the fix here is to point out what we mean by devsign

rtablada:
Makes some sense to be intentional, but at an outside glance it looks like it's a bad typo that really sticks out.

Maybe changing to explicitly saying "development and design"?

Inside lingo in a public post or notes seems like it's exclusionary :(

MelSumner:
Yes, it's intentional. I wouldn't consider it inside lingo since I picked it up in the developer community (I think it was someone at Microsoft?)

PR #3374, opened 2018-05-31
Release Ember.js Times Issue #49

jessica-jordan:
## What it does

To be released on June 1st, 2018.

Special Edition featuring @mansona and a look into the Guides app & its internals.

Feel free to add further topics to the comments below or mention any that you find interesting at #topic-embertimes

## Sources

screen shot 2018-06-01 at 00 56 42

## Related PRs

Once this is merged, we should also merge #3377 to highlight this post as the only recent Ember Times post.

Issue #49, opened 2012-02-10
Event links


jbrown:
I came across this today: http://www.meetup.com/Ember-SF/events/50058382/

Perhaps there should be a section, under community, that contains links to Ember-related events.

Thoughts?

**Update:** I found a Meetup list on the [ember.js wiki](https://github.com/emberjs/ember.js/wiki/Links). I've added the above link there.

PR #3377, opened 2018-05-31
blog(embertimes#48): remove #48 from recent posts

jessica-jordan:
## What it does

We're trying to only keep the latest edition of the Ember Times in the list of recent posts. Once edition No. 49 is out, we can remove No. 48 by removing the `Recent Posts` tag.

This PR removes the Issue #48 from the list of recent posts in favor of the latest edition #49.

## Related Issue(s)

Related to https://github.com/emberjs/website/pull/3374 and should be merged just after that one.


jessica-jordan:
@amyrlam Can you take a look?

PR #3384, opened 2018-06-07
contributing to the blog

tylerturdenpants:


amyrlam:
nice one 😄

made a couple copy edits. i think we've been doing one space after a period? fine with either, just should be consistent

love the smart quotes. i always have to google those and copy/paste them

PR #3388, opened 2018-06-08
Updated thoughtbot logo + `use` field

pablobm:
Updated logo for thoughtbot, as a company that uses Ember. Also added the "use" field since I was at it.


sivakumar-kailasam:
@pablobm the image defined in the yml file gets used by default. So please correct it to the one that you'd like to use.

pablobm:
Hi, I may be missing something, but I think you may be misreading the PR? I am providing:

* An image that replaces a previously existing one, with the same path. Therefore the path in the YAML file doesn't need to change.
* A new, higher resolution image, which from existing examples in the repo I understand is expected to appear in the same path, with a `@2x` suffix.

The "Files changed" section on the PR doesn't show image changes by default, so perhaps you missed that? When expanded, it looks like follows:

diff

Or perhaps it's me misunderstanding something here and I'm going to feel very embarrassed when you explain it back to me! 😳

sivakumar-kailasam:
@pablobm ah I should've clarified better. The 2x image isn't used we only use the one defined in the yml file. Unsure how the other images got into the system 😅 

ember_js_-_users

wifelette:
No embarrassment on any part, except maybe mine, because the instructions could always be better ;) 

Would be _nice_ to have it use the 2x images if they're there and when relevant, so if anyone feels like working on adding that, that'd be awesome too! And thanks for the update @pablobm <3

PR #3408, opened 2018-06-14
Add my google-site-verification

mansona:
This will allow me access to the Google Search Console as part of my SEO and Analytics efforts


sivakumar-kailasam:
@acorncom is there an official account that we can share instead of this?

acorncom:
@sivakumar-kailasam yes, looks like I added the line above Chris’, so I’ve got access there. Will work that out tomorrow when I’m at a computer

sivakumar-kailasam:
Great, closing this issue then

mansona:
Sure @acorncom we should chat about getting this sorted at some point soon 👍 

The reason I opened this PR was because I'm under the impression that "accounts" don't work the same way when it comes to the search console 🤔 that may have changed or I could be very mistaken 😂

PR #3412, opened 2018-06-17
Incremental updates to home page

MelSumner:
## What it does
If merged, this pull request would make the following changes: 

- updates "Get Started" to "Try it out!" 
- makes the h2 larger and tweaks the margins
- changes "More productive out of the box" to "Highly productive out of the box" 
- updated the text in the productivity section
- made these feature sections a little wider on larger screens (they still stack on smaller viewports)
- changed the "donate" button to "buy ember gear" (the header language and button CTA text was too discordant)

### Call to action
image

### Buy Ember Gear
image

### Productivity Features
image

This would also lead to future work- the intend is to add some additional information to each productivity feature, like this: 
update productivity section

### DIFF
--- a/source/about.html.erb
+++ b/source/about.html.erb
@@ -35,7 +35,7 @@ responsive: true
 <div class="call-to-action">
   <div class="call-to-action__box">
     <div class="call-to-action__top">
-      <h1>Get Started</h1>
+      <h1>Try it out!</h1>
       <div class="cta-copy">
         <p>Build your first Ember application with our 5 minute tutorial.</p>
       </div>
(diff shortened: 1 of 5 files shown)


MelSumner:
@rwjblue yay glad you like it

wifelette:
@MelSumner I think I see a small alignment issue? The copy under the handlebars is higher than the copy under the tower and Tomster.

Also, thoughts on where we could relocate the donate CTA to? We don't get _too_ much out of it, but we get _some_.

wifelette:
Actually, right now it says "buy gear" but still _points_ to the PayPal donate link :p So that'll need fixing either way.

PR #3417, opened 2018-06-20
Adds Native Class Constructor Update write-up to blog

Mi6u3l:
Adds Native Class Constructor Update write-up to blog

### DIFF
--- a/source/blog/2018-06-22-the-ember-times-issue-52.md
+++ b/source/blog/2018-06-22-the-ember-times-issue-52.md
@@ -13,7 +13,36 @@ Read either on the [Ember blog](https://www.emberjs.com/blog/2018/06/22/the-embe
 
 ---
 
-## [SECTION TITLE](section url)
+## [Native Class Constructor Update 🛠](https://github.com/emberjs/rfcs/pull/337)
+
+There is currently an open RFC proposing to change the behavior of EmberObject's constructor.
+
+Native class syntax with EmberObject has almost reached full feature parity, meaning soon Ember will be able to ship native classes.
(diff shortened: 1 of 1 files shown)


source/blog/2018-06-22-the-ember-times-issue-52.md
```diff
@@ -13,7 +13,36 @@ Read either on the [Ember blog](https://www.emberjs.com/blog/2018/06/22/the-embe
 
 ---
 
-## [SECTION TITLE](section url)
+## [Native Class Constructor Update 🛠](https://github.com/emberjs/rfcs/pull/337)
+
+There is currently an open RFC proposing to change the behavior of EmberObject's constructor.
```
jessica-jordan: We aimed to put anything that describes Ember objects, methods or similar things into code quotes (\`\`) in the past - I think it would make sense to put any mentions of EmberObject into ``, too  so, it'd become  `EmberObject`. 
What do you think?

source/blog/2018-06-22-the-ember-times-issue-52.md
```diff
@@ -13,7 +13,36 @@ Read either on the [Ember blog](https://www.emberjs.com/blog/2018/06/22/the-embe
...
+
+    return instance;
+  }
+}
+```
+
+This would assign the properties _after_ all of the class fields for any subclasses have been assigned.
+
+One thing worth mentioning is that EmberObject will likely be deprecated in the near future and that ideally for non-Ember classes (things that aren't Components, Services, etc.) users should drop EmberObject altogether and use native classes only.
+
```
jessica-jordan: I think it'd be cool to add a final call-to action as well, reminding people that they can comment on the RFC to help finalizing it - what do you think?

PR #3424, opened 2018-06-25
blog(embertimes#52): add more emphasis on the possible depr. of E.Obj

jessica-jordan:
## What it does

See also the wording from [the original RFC](https://github.com/pzuraq/emberjs-rfcs/blob/94b38d429eb2964fa86cd13bea6823a01b3ef68d/text/0000-native-class-constructor-update.md):

> One thing we should make clear is that EmberObject will likely be deprecated in the near future, and that ideally for non-Ember classes (things that aren't Components, Services, etc.) users should drop EmberObject altogether and use native classes only.

I changed it up a bit so it sounds less set into stone.


jessica-jordan:
<3

rwjblue:
Awesome, thank you @jessica-jordan!

PR #3425, opened 2018-06-26
Added section on SSD vs RAM Disk building

Alonski:
[Rendered](https://github.com/emberjs/website/pull/3425/files)


jessica-jordan:
@tylerturdenpants Would you like to review? I think you were considering writing this section as well and it'd be great to have your thoughts on this!

Alonski:
@tylerturdenpants, I added a few more sentences on the how. About this only being for *nix the author of the post is actually using Windows 😄 
And about how broccoli will achieve this:0 
```
[Tobias Bieniek](https://github.com/Turbo87) chimed in that soon Broccoli 2 will be fully integrated in Ember CLI and most platforms will get this speedup by default by using the system tmp folder instead of the current `in-project` tmp folder.
```

PR #3447, opened 2018-07-10
Reduce line-height of title

samselikoff:
Line height and text size are typically inverses, so I'm proposing we take down the line-height of the title a bit.

Happy to provide some material on this if anyone would like!


acorncom:
Before:
before

After:
after

samselikoff:
You're welcome! ❤️

PR #3458, opened 2018-07-13
Add blog post announcing requirement of 2-factor auth

kategengler:
Add a blog post to announce our policy on 2factor auth for our github orgs


kategengler:
I took the org names as they are on GitHub, maybe we should standardize

PR #3463, opened 2018-07-18
Update broken link

amyrlam:
This Goodbits link has a different pattern than https://the-emberjs-times.ongoodbits.com/2018/07/06/the-ember-times-issue-54 or https://the-emberjs-times.ongoodbits.com/2018/06/29/the-ember-times-issue-53. Should we fix here or on Goodbits?


Alonski:
Woops. I think it would make sense to fix on goodbits so emails will be correct. Need to watch out for this next time.

PR #3464, opened 2018-07-19
Amy supershort writeup for Ember Times 56

amyrlam:
First stab - may try to flesh out tmrw. Any feedback let me know!

I found it tough to say much about the podcast but could summarize the RFCs if needed. I also wasn't sure if it is indeed "splatributes"?

There is some double bracket curly syntax in here that may need to be removed also... will comment inline too.

### DIFF
--- a/source/blog/2018-07-20-the-ember-times-issue-56.md
+++ b/source/blog/2018-07-20-the-ember-times-issue-56.md
@@ -17,7 +17,6 @@ Check out what's been going in Emberland 🐹 these days:
 ## IDEAS
 - Edward Faulkner@eaf4
 Would you like to use dynamic import() to lazy-load deps into your Ember app right now? Here's a video showing how: https://eaf4.com/dynamic-import-into-your-ember-app/
-- ember weekly podcast
 - https://twitter.com/mixonic/status/1017487614121684992
 - https://twitter.com/oligriffiths/status/1019257665925648390?s=19
 ---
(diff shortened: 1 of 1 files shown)


source/blog/2018-07-20-the-ember-times-issue-56.md
```diff
@@ -38,11 +38,11 @@ Would you like to use dynamic import() to lazy-load deps into your Ember app rig
 
 ## [SECTION TITLE](#section-url)
 
-
 ---
 
-## [SECTION TITLE](#section-url)
+## [RFC Roundup with @rwjblue](https://emberweekend.com/episodes/rfc-roundup-with-rwjblue)
 
+The Ember Weekend podcast is back on the mic! Want to hear more about some recent RFCs? @rwjblue discusses [RFC #276: Named Arguments Syntax (`{{@foo}}`)](https://github.com/emberjs/rfcs/blob/master/text/0276-named-args.md), [RFC #213: Custom Components API](https://github.com/emberjs/rfcs/pull/213), and [RFC #311: Introduce `<AngleBracketInvocationSyntax />`](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md) in [Ember Weekend's latest podcast](https://emberweekend.com/episodes/rfc-roundup-with-rwjblue). Entertaining and informative! TIL "splatributes". Subscribe to Ember Weekend [here](https://itunes.apple.com/us/podcast/ember-weekend/id981719021) on iTunes. 🎙
```
amyrlam: Would it be better to not write `{{@foo}}` at all with Goodbits?

jessica-jordan: Yes, good point, I think it makes it easier to not forget about removing it in goodbits and I think in this case you could also be able to convey the message by writing `@foo` on its own. What do you or @tylerturdenpants think?

tylerturdenpants: I prefer clarity over assumption. As for goodbits, I think we may need to add it the template to remember to fix it before submitting it.

source/blog/2018-07-20-the-ember-times-issue-56.md
```diff
@@ -66,15 +66,8 @@ Would you like to use dynamic import() to lazy-load deps into your Ember app rig
...
-## [The Ember Times is What We Make It 🙌](https://embercommunity.slack.com/messages/C8P6UPWNN/)
-
-The **Ember Times** is a **weekly news editorial** featuring all the new things that are going on in Emberland.
-[Subscribe to our e-mail newsletter](https://the-emberjs-times.ongoodbits.com/) to get the next edition **right to your inbox**.
-And if you've always wanted to be an OSS journalist yourself,
-drop by [#topic-embertimes](https://embercommunity.slack.com/messages/C8P6UPWNN/)
-on the Ember Community [Slack Chat](https://ember-community-slackin.herokuapp.com/)
-and **write** the next edition of the Ember Times **together with us**!
-
+Have a suggestion for next week's Ember Times? Want to write for us? Pop into [#topic-embertimes](https://embercommunity.slack.com/messages/C8P6UPWNN/)
```
tylerturdenpants: @amyrlam Thanks for this update

PR #3479, opened 2018-07-27
Fixes logo on subpages

jelhan:
Logo was using a relative URL and therefore not being available on any subpage (expect for API docs which is generating header on it's own). This was introduced in 9278d2381788dc607fcef62efe6e92c2953e7ae6.


amyrlam:
👍 many thanks @jelhan!

PR #3519, opened 2018-08-21
For EmberTimes#61: Add Modifiers RFC + Readers' Question

jessica-jordan:
## What it does

Adds a short write up on the Modifiers RFC and the almost finished intro this week's Readers' Question.


jessica-jordan:
Thank you for checking, @amyrlam ✨

PR #3525, opened 2018-08-23
Remove InQuicker from Users

kpfefferle:
## What it does

Removes InQuicker from the list of Ember Users

## Sources

I added InQuicker to the Users list when I worked there and we were actively advancing Ember development within the product and company. Sadly, our team has disbanded and there is no longer active Ember development going on within InQuicker.


kpfefferle:
@sivakumar-kailasam Done, thanks for the direction!

sivakumar-kailasam:
Thanks @kpfefferle

PR #3532, opened 2018-08-28
Added section on angle bracket components

kennethlarsen:


source/blog/2018-08-27-ember-3-4-released.md
```diff
@@ -24,8 +24,40 @@ Ember.js is the core framework for building ambitious web applications.
 
 Ember.js 3.4 is an incremental, backwards compatible release of Ember with bugfixes, performance improvements, and minor deprecations. There is one (1) new feature, three (3) deprecations, and eight (8) bugfixes in this version.
 
-#### New Features (X)
+#### New Features (2)
+Angle Bracket Invocation (1 of 2)
+
+In Ember 3.4 it is now possible to invoke angle bracket components. This means that you're now able to replace the classic invocation syntax:
```
CodingItWrong: I have a bit of concern that the phrase "angle bracket components" might lead people to think that it's a different kind of *component*, rather than just a different invocation syntax. I think this is based on the talk of allowing the use of "Glimmer components" in Ember (which I still don't quite understand tbh). Maybe this is being overly cautious, since the code example shows the same component with two different syntaxes. But angle bracket syntax is a big enough change that caution can be good.

How about this? "In Ember 3.4 it is now possible to invoke components using angle brackets."

source/blog/2018-08-27-ember-3-4-released.md
```diff
@@ -24,8 +24,40 @@ Ember.js is the core framework for building ambitious web applications.
...
+<SuperSelect @selected={{this.user.country}} as |Option|>
+  {{#each this.availableCountries as |country|}}
+    <Option @value={{country}}>{{country.name}}</Option>
+  {{/each}}
+</SuperSelect>
+```
+
+It's important to note that the classic invocation syntax isn't deprecated in favour of this new feature. You're free to still use the classic invocation syntax but the angle bracket components do have a few advantages.
+
+By using angle bracket components you introduce more clarity to your templates. The main advantage of the angle bracket syntax is clarity. Because component invocation is often encapsulating important pieces of UI, a dedicated syntax would help visually distinguish them from other handlebars constructs, such as control flow and dynamic values. This can be seen in the example shown above – the angle bracket syntax made it very easy to see the component invocations as well as the `{{#each}}` loop, especially with syntax highlight.
```
CodingItWrong: As per above, "By using angle bracket syntax for components you introduce…"

source/blog/2018-08-27-ember-3-4-released.md
kennethlarsen: Yeah I think that's a good point, thanks!

source/blog/2018-08-27-ember-3-4-released.md
```diff
@@ -24,8 +24,40 @@ Ember.js is the core framework for building ambitious web applications.
...
+Angle Bracket Invocation (1 of 2)
+
+In Ember 3.4 it is now possible to invoke angle bracket components. This means that you're now able to replace the classic invocation syntax:
+
+```hbs
+{{site-header user=this.user class=(if this.user.isAdmin "admin")}}
+
+{{#super-select selected=this.user.country as |s|}}
+  {{#each this.availableCountries as |country|}}
+    {{#s.option value=country}}{{country.name}}{{/s.option}}
```
chrisrng: Thoughts on making the classic syntax example match angle bracket?

`{{#super-select selected=this.user.country as |s|}}`

vs having option instead of "s"

`{{#super-select selected=this.user.country as |option|}}`

kennethlarsen: Good catch, yeah that's probably more helpful that way. Thanks!

source/blog/2018-08-27-ember-3-4-released.md
Alonski: I actually think that Angle Bracket Components is the _official_ name for this no?

CodingItWrong: I just took a quick look through the RFC:

https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md

"Angle bracket invocation" and "angle bracket invocation syntax" seem to be the most common. The only reference I could find to "angle bracket components" was a link to an earlier RFC, and from the description it sounded like that RFC *did* change the nature of the components, whereas this RFC only changes the invocation syntax.

rwjblue: @codingitwrong is correct here, the feature added is “angle bracket invocation”. It’s _super_ important to make it clear that the invocation syntax **does not** change anything about the component itself (base class, behaviors, etc).

kennethlarsen: @rwjblue Thank you for the clarification!

PR #3541, opened 2018-08-31
Update mmun's teams list

mmun:
:v:


mmun:
@acorncom :)

Become an alumni is a reflection of my activity level on core and not of my passion for Ember. I'll continue contributing and being a part of the community as I have been.

PR #3542, opened 2018-09-01
Adding Newsletter tag to Browse by Subject

Alonski:
## What it does

This adds the Newsletter tag link under Browse by Subject

## Sources

Adding here:
image


Alonski:
@amyrlam Good catch. Fixed

PR #3550, opened 2018-09-06
[Ember Times] Update embertimes-template.md

amyrlam:
This is repeated in every blog post on emberjs.com/blog. Put in Goodbits only? See #support-ember-times on Discord. cc @cah-danmonroe / @danmonroe

### DIFF
--- a/source/blog/embertimes-template.md
+++ b/source/blog/embertimes-template.md
@@ -81,7 +81,7 @@ Read either on the [Ember blog](https://emberjs.com/blog/tags/newsletter.html) o
 
 ---
 
-Want to write for the Ember Times? Have a suggestion for next week's issue? Join us at #support-ember-times on the [Ember Community Discord](https://discordapp.com/invite/zT3asNS) or tweet us [@embertimes](https://twitter.com/embertimes) on Twitter.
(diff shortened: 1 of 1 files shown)


amyrlam:
Update https://github.com/emberjs/website/pull/3540 if ok with this!

source/blog/embertimes-template.md
```diff
@@ -8,7 +8,9 @@ responsive: true
 
 <SAYING-HELLO-IN-YOUR-FAVORITE-LANGUAGE> Emberistas! 🐹
 
+<goodbits-only>
```
jessica-jordan: Does this one work like a HTML comment on the page?

amyrlam: No...just wasn't sure how to specify "template for Goodbits, but not the blog". Is a template on goodbits.com only a better way?

jessica-jordan: Ah okay I see! Yes sure, I like your idea to just add it to goodbits - once we have it in one edition we can just copy over the template each week. Would you like to remove this section from the blog template then?

amyrlam: Removed it here, so this PR should be good to go.

I am not 100% sure what is the best way to create a template in Goodbits / create a new Goodbits (the couple times I did it from scratch it was a little time consuming). Maybe we could pair on Goodbits / take a video of you Goodbits-ing if possible? I can also help more in Goodbits so know this process better...

PR #3562, opened 2018-09-14
[Ember Times] Amy writeup for issue 64

amyrlam:


amyrlam:
It seems like we can use https://capitalizemytitle.com/# for any section headers we aren't sure about, if people are 🆒 with it!

jessica-jordan:
> It seems like we can use https://capitalizemytitle.com/# for any section headers we aren't sure about, if people are 🆒 with it!

@amyrlam That's a cool idea, I didn't know this one - I'll check it out!

PR #3566, opened 2018-09-19
blog(embertimes65): sparkles-component 1.1.0

chrisrng:
[rendered](https://github.com/chrisrng/website/blob/7351fbab192c79aaf17abccf767c82e7fc4215cb/source/blog/2018-09-21-the-ember-times-issue-65.md#published-sparkles-component-110-)

PR #3567, opened 2018-09-20
[Ember Times] Add alex linter ignores

amyrlam:
Relates to: https://github.com/emberjs/website/pull/3566

The build for https://github.com/emberjs/website/pull/3566 passed, but when it was merged into https://github.com/emberjs/website/pull/3563 it failed


amyrlam:
I found it super confusing that the individual build didn't fail (and the builds are quite delayed since open source), but since we are just getting started with alex, seems ok! 

I have alex working locally now... 

(am confused by the VSCode plugin)

PR #3570, opened 2018-09-20
[Ember Times] Update Ember Times template

amyrlam:
Originally we removed this from the blog, and just put it in the Goodbits, based on reader feedback. 

What do you think about moving this to the "Want to write for the Ember Times?" section, since it's not necessarily a way to hook the reader? Then could be back in both(?) or just the Goodbits?

This is minutiae so just tagging you two for now since don't want to overload the Times team with PR's tonite. That being said, anyone's feedback is welcome!

### DIFF
--- a/source/blog/embertimes-template.md
+++ b/source/blog/embertimes-template.md
@@ -77,7 +77,9 @@ responsive: true
 
 ---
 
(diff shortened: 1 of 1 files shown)


tylerturdenpants:
I’d like to see it in both places.

amyrlam:
🆒 let's try it, can iterate if need be. Seems less "why is this repeating every week" if not at the top

PR #3588, opened 2018-09-28
oops (typo fix) for ET 66

NullVoxPopuli:
strucure => structure


amyrlam:
i did that here https://github.com/emberjs/website/pull/3587 with another change, so if you don't mind I'll close this one

PR #3565, opened 2018-09-18
Additions to meetup branding

jeffdaley:
- New branding and assets for meetups
- Updated brand guidelines
- Updated meetup-assets page

PR #3598, opened 2018-10-03
Updates to Meetup branding and brand guidelines

amyrlam:
- New branding and assets for meetups
- Updated brand guidelines
- Updated meetup-assets page

## Sources
This PR will close / take the place of https://github.com/emberjs/website/pull/3565 once we confirm that it is indeed the same content, minus the `Gemfile`.  Will also close / take the place of https://github.com/emberjs/website/pull/3601.


amyrlam:
@jeffdaley I had a lot of trouble rebasing your branch so I created a new one (as joint committers). 

I _think_ I got everything else from https://github.com/emberjs/website/pull/3565, minus the `Gemfile` which I don't think we need to commit. 

There are now a few sass errors that we need to debug...

source/stylesheets/pages/meetup-assets.css.scss
```diff
@@ -1,3 +1,13 @@
+.meetup-assets {
+  .example-meetups-logo {
+    float: right;
+    @include margin(null null 15px 15px);
```
amyrlam: @jeffdaley If you run `bundle exec middleman build` locally you'll get:
 
```
bundler: failed to load command: middleman (/usr/local/bin/middleman)
Sass::SyntaxError: Undefined mixin 'margin'.
```

Do we need a `@mixin margin`?

jeffdaley: I thought that would work with the Bourbon gem. Guess not!

I rewrote in plain css and resubmitted.

amyrlam:
Let's use #3601 instead

PR #3621, opened 2018-10-09
Corrected link to vanilla codesandbox example

yohanmishkin:


amyrlam:
Thanks for the fix! 💯 

Still not sure what's up with Travis but was talking to @sivakumar-kailasam and this seems safe to merge.

amyrlam:
`master` passed, yay

PR #3637, opened 2018-10-14
Improve Who uses ember on homepage

### LINKED ISSUE #3134: Improve the "who uses Ember" carousel on the home page.
There could be two approaches here: 
1. update the slider used
2. remove that slider (and customizations) and replace it with something different. 

Either approach: 
- should be responsive/device-agnostic
- should complement the current site
- should be accessible (A11y)

spruce:
## What it does
Updates the slider to fix the jumping when changing over the end (#3346).
Also removes the slider controls as suggested in https://github.com/emberjs/website/issues/3134#issuecomment-377826001 .

## Related Issue(s)

Closes #3134


amyrlam:
@spruce :wave: Sorry for the delay in responding to this!

It _seems_ like this has been covered by https://github.com/emberjs/website/pull/3648 and with the current state of the webpage (below), this PR could be closed. 

image

Let me know what you think - sorry for the confusion!

spruce:
This seems to be the better solution. I like the new design.

PR #3644, opened 2018-10-19
Short writeup on EmberConf brainstorm

amyrlam:
Also, added authors and cleaned up the branch

Still need:
- [x] hello in language we haven't used...?
- [ ] intro summarizing content


kennethlarsen:
Here's a hello I think we haven't used: "Goddaw"

Alonski:
Also, this will probably conflict with my PR. The one that gets merged second will need to have conflicts fixed.

PR #3649, opened 2018-10-20
Disable search temporarily

sivakumar-kailasam:
Our API & guides apps have functional search, but until we can fix the search on the rest of the website its probably better to hide the search bar so that users don't feel that its broken.

### DIFF
--- a/source/_header.html.erb
+++ b/source/_header.html.erb
@@ -66,7 +66,7 @@
       </ul>
      </li>
    </ul>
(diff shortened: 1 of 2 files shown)


amyrlam:
Talked to Siva on Discord. 

It takes a bit of time for the search "popover" to appear. Add a loading indicator?

When you press "return", search results go away, so it _seems_ like it's not working, but it does if you _do not_ push "return".

Going to look into what's going on / issue-ify these.

computed

PR #3777, opened 2019-01-25
Angle brackets and contributors for Ember Times

amyrlam:
let me know if any feedback!

### DIFF
--- a/source/blog/2019-01-25-the-ember-times-issue-82.md
+++ b/source/blog/2019-01-25-the-ember-times-issue-82.md
@@ -14,11 +14,11 @@ responsive: true
 
 ## [SECTION TITLE](#section-url)
 
-
 ---
 
-## [SECTION TITLE](#section-url)
+## [Bringing Angle Brackets to the Guides 🥳](https://github.com/ember-learn/guides-source/issues/139)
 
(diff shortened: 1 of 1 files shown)


amyrlam:
Note: it seems safe to say "is coming soon" because https://guides.emberjs.com/release/components/block-params/ still reflects the old although https://github.com/ember-learn/guides-source/pull/357 is merged (guessing the deployments for this repo are not automatic intentionally?)

source/blog/2019-01-25-the-ember-times-issue-82.md
```diff
@@ -14,11 +14,11 @@ responsive: true
 
 ## [SECTION TITLE](#section-url)
 
-
 ---
 
-## [SECTION TITLE](#section-url)
+## [Bringing Angle Brackets to the Guides 🥳](https://github.com/ember-learn/guides-source/issues/139)
```
kennethlarsen: Not sure if only on my laptop but the emoji is just a square for me?

kennethlarsen:
@amyrlam Yes it is intentional that it isn't deployed yet so I think saying "coming soon" is the right approach

source/blog/2019-01-25-the-ember-times-issue-82.md
amyrlam: Thanks - it's the party "face" but maybe it's too new of an emoji, changed to the older :tada:

PR #3793, opened 2019-02-01
Update 2019-01-26-emberjs-native-class-update-2019-edition.md

amyrlam:
## What it does
@pzuraq hey great article! I was reading this for the Ember Times and made some copy edits here.

## Aside

To fix later...I'm not sure why this is showing `1.` for both:

image


pzuraq:
Thanks for catching those! And yeah, that must be some kind of markdown parser bug, seems strange

amyrlam:
👍 made an issue: https://github.com/emberjs/website/issues/3796

PR #3797, opened 2019-02-01
Non-update to template

amyrlam:
Attempt to fix Travis woes


amyrlam:
This is no longer needed, was a test to unblock CI woes on Friday

PR #3852, opened 2019-03-06
Remove 3.9 deprecations

kennethlarsen:
## What it does
It seems that four deprecations were wrongfully tagged as `3.8` instead of `3.9` so I have updated the release post to reflect reality.


tavosansal:
For what is worth... the header here still says `(5)` https://emberjs.com/blog/2019/02/27/ember-3-8-released.html#toc_deprecations-5

Also does that mean that this sidebar should also be updated?
image

It is pretty confusing right now. I am not sure if these deprecations apply to 3.8 or not

